PR #6, opened 2026-06-27
fix(cloud): send a browser User-Agent so Orbit's WAF doesn't 403 setup

### LINKED ISSUE #1: Auth failure
I'm getting an auth failure - possibly related to changes made by Orbit today, as the cloud integration is also having problems.  This may be relevant, assuming you're hitting the same API to get the tokens.

https://github.com/sebr/bhyve-home-assistant/issues/427

exactmike:
## Summary

Setup fails with **"invalid email or password"** even when the credentials are
correct. The real cause isn't auth at all: Orbit's API gateway (WAF) returns
**HTTP 403** to any request whose `User-Agent` contains `HomeAssistant` — which
is exactly what Home Assistant's shared aiohttp session
(`async_get_clientsession`) stamps on every outbound request by default. So
*every* cloud call the integration makes at setup is rejected before it's even
evaluated, and `config_flow` mislabels the 403 as bad credentials.

This is the same ecosystem-wide event that hit the cloud-based B-Hyve
integration ([sebr/bhyve-home-assistant#427](https://github.com/sebr/bhyve-home-assistant/issues/427), fixed there in 4.1.1).

## Root cause

`OrbitCloudClient._headers()` sent only `{"orbit-app-id": ...}`, so requests
inherited HA's default `User-Agent: HomeAssistant/<ver> aiohttp/<ver>`. The WAF
403s on that substring. Verified against the live API with dummy credentials:

| User-Agent | Result |
|---|---|
| `HomeAssistant/... aiohttp/...` | **403 Forbidden** (WAF block) |
| `python-requests/...` | 401 (reaches the app — normal "bad creds") |
| `Mozilla/5.0 ...` | 401 (reaches the app) |
| *(empty)* | 401 (reaches the app) |

The 401s prove the request *reaches* the application layer once the UA isn't
`HomeAssistant`; only the WAF-403 path is the bug.

## Fix

Send a browser `User-Agent` (and a normal `Accept`) on cloud calls, via a new
`const.CLOUD_USER_AGENT`:

```python
def _headers(self, *, include_auth: bool = True) -> dict[str, str]:
    h = {
        "orbit-app-id": CLOUD_APP_ID,
        "User-Agent": CLOUD_USER_AGENT,
        "Accept": "application/json, text/plain, */*",
    }
    if include_auth and self._token:
        h["orbit-api-key"] = self._token
    return h
```

## Scope & backwards compatibility

- **One place, additive.** Only the cloud REST `_headers()` changes.
- This integration contacts the cloud **only at setup** (no websocket, no
  background polling), so the change touches the setup path exclusively and has
  no effect on the BLE/runtime side.
- No behavior change for any already-working account beyond no longer being
  WAF-blocked.

## Testing

- Before: config flow fails immediately with "invalid email or password" on a
  known-good account.
- After: config flow logs in, discovers all devices on the account, and fetches
  each device's BLE network key — setup completes and entities appear.

## Related / follow-up

A second, independent issue can also surface the same "invalid email or
password" message on **newer-schema accounts** (key fetch aborting on a 401 from
`/meshes`). That's addressed in a separate PR so this universal WAF fix can land
on its own.

Fixes #1

### DIFF
--- a/custom_components/orbit_bhyve/cloud.py
+++ b/custom_components/orbit_bhyve/cloud.py
@@ -15,6 +15,7 @@ from .const import (
     CLOUD_APP_ID,
     CLOUD_KEY_FIELDS,
     CLOUD_KEY_PATHS,
+    CLOUD_USER_AGENT,
 )
 
 _LOGGER = logging.getLogger(__name__)
@@ -49,7 +50,17 @@ class OrbitCloudClient:
         return self._user_id
 
     def _headers(self, *, include_auth: bool = True) -> dict[str, str]:
-        h = {"orbit-app-id": CLOUD_APP_ID}
+        # Orbit's WAF returns 403 to requests whose User-Agent contains
+        # "HomeAssistant" — which is exactly what HA's shared aiohttp session
+        # (async_get_clientsession) stamps on every request. Without an
+        # override, every cloud call here is forbidden and config_flow
+        # mislabels it "invalid email or password". Send a browser User-Agent
+        # (matching the wider B-Hyve HA ecosystem's fix) so setup works.
+        h = {
+            "orbit-app-id": CLOUD_APP_ID,
+            "User-Agent": CLOUD_USER_AGENT,
+            "Accept": "application/json, text/plain, */*",
+        }
         if include_auth and self._token:
             h["orbit-api-key"] = self._token
         return h
--- a/custom_components/orbit_bhyve/const.py
+++ b/custom_components/orbit_bhyve/const.py
@@ -20,6 +20,13 @@ SERVICE_UUID = "0000fe32-0000-1000-8000-00805f9b34fb"
 # Cloud API.
 CLOUD_API_BASE = "https://api.orbitbhyve.com/v1"
 CLOUD_APP_ID = "Bhyve-App"
+# Orbit's WAF 403s any request whose User-Agent contains "HomeAssistant"
+# (which HA's shared aiohttp session sets by default). Send a browser UA on
+# cloud calls so setup isn't rejected. See sebr/bhyve-home-assistant#427.
+CLOUD_USER_AGENT = (
+    "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) "
+    "Chrome/72.0.3626.81 Safari/537.36"
+)
 CLOUD_KEY_PATHS = (
     "/meshes/{mesh_id}",
     "/network_topologies/{mesh_id}",


exactmike:
Credit where due — this overlaps with prior art on the same root cause: @stuartdenne's #5 and the same fix in knobunc/ha-orbit-bhyve-ble (`9014609`) both identified Orbit's WAF returning 403 to the default `HomeAssistant` User-Agent.

What this PR adds is **live verification**:
- Against the Orbit API directly: UA `HomeAssistant/... aiohttp/...` → **403**, while `python-requests`/`Mozilla`/empty → 401 — i.e. the request only reaches the application layer (normal "bad creds") once the UA isn't `HomeAssistant`.
- End-to-end: on a previously-broken auth state, setup now completes and discovers a full account (4× HT25G2 valves + an HT34A XD + hub).

@stuartdenne — we've been doing a fair amount of BLE-first work on this integration (hardware-verified actuation + RX telemetry across a 6-device fleet, plus a standalone CLI/RE toolchain for live testing). Would love to compare notes and avoid duplicating effort. Happy to align on this cloud-auth fix however @ljmerza prefers.

stuartdenne:
@exactmike — glad you pinged. Nice work on the live verification. We landed the same trailer formula and the 2³² counter-wrap independently, which is a good sign we've both got the cipher right.

Lay of the land as I see it: this has fragmented into ~5 repos — wxfield (the original) → ljmerza → knobunc → my fork, plus your wxfield fork — and every upstream maintainer has gone quiet. The cloud-auth fix alone is now sitting unmerged in three places (your #6, my closed #5 here, knobunc `9014609`). You and I look like the only two actively shipping, so I'd rather join forces than keep duplicating.

What I've got that complements your RX work:

- 5× HT25-0000 hose-taps (fw `0085`) hardware-verified — actuation fixed with per-device mesh-ID addressing (replaced an upstream hardcoded-address path that only worked for a single unit), plus re-init before every actuation with status-poll confirmation and one retry (the HT25 ACKs but silently no-ops a watering command on a stale pooled session — no solenoid), and a bounded handshake + whole-open retry for marginal links over ESPHome BT proxies.
- An HT34-0001 4-port (fw `0058`) on the protobuf/CRC path — start/stop verified against decoded watering-state, plus battery decode. Looks like your 4-port is the HT34A/fw `0107` and your hose valves are HT25G2/fw `111`, so between us we cover different hardware+firmware in both families.
- Packaged as a full HACS integration — valve-per-station entities, battery/RSSI sensors, default-duration number, sync button, debug probe services.

Where you're ahead of me: I only decode the single watering-status byte to confirm a command took — I don't have a general device→host decode. Your RX keystream / status catalog is exactly the gap on my side, so merging the two efforts is the obvious win.

On where to consolidate: rather than keep PRing into dormant repos, I think we pick one actively co-maintained home and treat the rest as upstreams to credit. Two options I see — (1) revive the original wxfield if its owner's up for making it canonical and adding us both as maintainers (fits your `Orbit_B-Hyve_BLE` rename idea), or (2) stand up a fresh co-owned repo and move on. I lean toward whichever gets us merge control fastest. What's your preference?

Happy to take the detail off this thread — your repo's Discussions, a new issue, or email, whatever suits.

exactmike:
@stuartdenne — yes, let's join forces. Your read matches mine and the split is genuinely complementary: you've got the fw0085 mesh reliability, the HT34-0001 (fw0058) coverage, and the HACS packaging dialed in; we've got the RX keystream solve and a full device→host status/capability decode catalog — the gap you flagged on your side. Happy to share that now regardless of how we organize: it's written up on the `integration` branch of our fork (local-ha) as `docs/ble_protocol.md` + `protobuf/orbit_ble.proto`, yours to use.

One reframe on *where* to consolidate: when you wrote that, the upstreams looked dormant — but @ljmerza has merged five PRs in the last couple of days (ours included) and has been turning things around in hours. So rather than stand up a new repo, I'd rather we do this **here**, properly multi-maintained.

@ljmerza — would you be open to adding stuartdenne and me as collaborators, and/or moving the repo into a shared org? Entirely your call and you'd stay in the driver's seat — the goal is just resilience. This integration has fragmented into ~5 forks largely because each rode on a single maintainer; three of us shipping under one roof with a clear division of labor (RX/decode + a live-hardware CLI test harness on our side, mesh reliability + packaging on stuartdenne's) is the durable fix. If a shared org isn't your thing, no worries — stuartdenne and I have a co-owned-repo fallback — but I'd much rather build on what you've already got here.

Either way, a couple of overlaps to deconflict so you don't get duplicate PRs: stuartdenne and I both have a bounded-handshake + connect-retry change, and the fw0085 mesh path now has work from stuartdenne, @FabsFabios, and us — worth sorting who owns each before sending more.

Happy to take the org/logistics detail wherever suits you — here, a Discussion, an issue, or email.

ljmerza:
It's just AI talking to AI in here. Wild for strangers to boldly ask to be maintainers to a new project youve contributed once to. You're free to open PRs or try your luck with a fork but this project has already been accepted into HACS so doubt they'd be interested in adding a bunch of forks.

exactmike:
Thanks for the response. I appreciate and accept your position. Apologies for overstepping. I'm a real person who has admittedly been using AI to assist. My intention is not to have many forks, other than for the purpose of developing to contribute here.  Glad for the acceptance of your repo into HACS and just wanting to continue contributing. Also interested in reducing duplicate effort in the community - stuartdenne and I have worked on some overlapping ground.  I will keep opening PRs here and greatly appreciate your responsiveness. Let me know if there's anything different you would like to see with future PR requests.

stuartdenne:
In HACS now?  Cool.

@exactmike / @ljmerza - thanks, looks like the infinite-claude-forks problem is solved.  The resiliency I added can probably be improved further; but then so could Orbit's gateway implementation of it!  I'll likely move on to flow metering next (good for leak detection), although that's going to need some btsnooping.

PR #18, opened 2026-07-05
docs: BLE transport-reliability & hardware-verified behavior (Gen2 + XD)

exactmike:
## docs: BLE transport-reliability & hardware-verified behavior (Gen2 + XD)

Documentation only — **no code changes.** Adds one focused doc,
`docs/ble-reliability-and-behavior.md`, covering the **operational** layer of the Gen2 (HT25G2) and
XD (HT34A) BLE protocol: the transport-reliability behavior and the fine-grained, hardware-verified
status-decode semantics needed to control these valves reliably over an ESPHome/BlueZ proxy.

It is **deliberately complementary** to the existing protocol documentation — it does **not** restate
the GATT layout, cipher, handshake, framing, or message schema (already well covered; see *Prior art*
in the doc). It adds the layer on top that those don't:

**Transport reliability** (not documented elsewhere):
- single-BLE-session limit and why an ephemeral connect-per-operation model beats a pooled session;
- solicited vs. unsolicited RX (need a `#15{}` elicitor to read state while active);
- **duplicate-RX-frame CTR desync** and the byte-identical-dedup fix;
- a **persistent `#57` flow subscription starving `#16`** (+ the `#57{#1=0}` fix) and `#59`-desync recovery;
- firmware that reports a **static `remaining`** and the wall-clock drift-guard;
- CTR-desync self-heal.

**Hardware-verified status-decode semantics:**
- run progress `#16.#6/#7` with the **Gen2↔XD layout flip** (`#16.#7.#6` vs `#16.#6.#7`);
- **active zone** at `#16.#2.#2.#3.#1` (the only running-station report on the XD);
- STOP returns a **bare `#30` ack** → confirm via a follow-up `#15{}`;
- rain delay honors an **absolute `#3` expiry** (anchor to the device clock `#7`) + the bare-clear `#4`-absent gotcha;
- flow `#57`/`#59` is **Gen2-only**, and `#59.#3` is a **cumulative counter, not a rate** (`#59.#1` = "flowing", not "valve open").

### Provenance, safety & credits
- Validated on hardware the authors own — Gen2 fw `0111`, XD HT34A fw `0107` — over ESPHome proxies;
  each item tagged **[HW-verified]** / **[capture-decoded]**. No vendor firmware/app code; no secrets.
- Credits **wxfield** (originator of the RE — cipher, framing, initial schema; MIT) and **knobunc**
  (whose `PROTOCOL_SPEC.md` + `protobuf_schema.json` are the structural reference this builds on).

[...]

Issue #16, opened 2026-07-02
Hardware verification of v2.1.0/v2.1.1 (HT25G2 fw0111 + HT34A fw0107)


exactmike:
You flagged several v2.1.0/v2.1.1 changes as untested on hardware. I deployed **exactly this release** (v2.1.1, unmodified) to a live Home Assistant instance and drove those paths on real devices, over an ESPHome Bluetooth proxy. Everything works. Details + evidence below.

**Hardware under test (all via ESPHome BT proxy `athom-rf-ir-remote-*`):**
- 4× Gen2 **HT25G2-0001**, firmware **0111** (single-station hose-tap valves)
- 1× **HT34A-0001**, firmware **0107** (4-port XD)

### 1. Bounded handshake + retry + write-ack cap — `97cb69b`  ✅
Clean connect + handshake on both device families over the proxy, no ~30 s wedge:
- XD: connect 1.9 s → handshake 0.4 s (`handshake ok, iv=8aff34c4… tx_ctr=0x3d6eb008`)
- Gen2: connect 3.7 s → handshake clean (`iv=d09b87c2… tx_ctr=0xae118408`)

The write-ack cap behaves correctly: commands return promptly and the notification drain is the real ack, exactly as intended (the proxy never relays the ATT write response).

### 2. Battery / RSSI / Connected / Watering sensors — `f1084f7`  ✅
All 5 devices populate live and per-device-distinct; Connected/Watering transitions track real state (Connected flips on at the connect timestamp; Watering flips on/off with actuation):

| Device | Model / fw | Battery | Voltage | RSSI |
|---|---|---|---|---|
| BT4ValveXD01 | HT34A / 0107 | 89 % | 2932 mV | −48 dBm |
| BTValve01 | HT25G2 / 0111 | 41 % | 2626 mV | −61 dBm |
| BTValve02 | HT25G2 / 0111 | 43 % | — | −56 dBm |
| BTValve03 | HT25G2 / 0111 | 50 % | — | −36 dBm |
| BTValve04 | HT25G2 / 0111 | 31 % | 2572 mV | — |

[...]

Issue #13, opened 2026-07-01
no device class for hardware='HT32A-0001'


jameshallam:
I just installed Version 2.1.0, and was able to authenticate. I have three 2-port HT32A timers, but none of them appear under the account listing once the authentication is complete. 

Here is the full error log:

```
This error originated from a custom integration.

Logger: custom_components.orbit_bhyve
Source: custom_components/orbit_bhyve/__init__.py:77
Integration: Orbit B-Hyve BLE (documentation, issues)
First occurred: 11:33:33 PM (9 occurrences)
Last logged: 11:35:15 PM

Deck Water Timer: no device class for hardware='HT32A-0001' firmware='0107' — skipping
Front Porch Water Timer: no device class for hardware='HT32A-0001' firmware='0107' — skipping
Greenhouse Water Timer: no device class for hardware='HT32A-0001' firmware='0107' — skipping
```

Is it possible to add support for the HT32A along with the HT34A?

Thanks for all of the recent effort on this project.

ljmerza:
Hey without the device here to decode the Bluetooth connection, there's not a lot I can do. If you can figure it out, you can enable ble snoop, connect and turn on and off the device, then download the interest bug report and send it to me.

ljmerza:
Support for the HT32A is in **v2.1.1** — just released: https://github.com/ljmerza/orbit-bhyve-ble/releases/tag/v2.1.1

The HT32A-0001 turns out to be the 2-port sibling of the 4-port HT34A: same firmware (0107) and the same protobuf BLE protocol, just fewer stations. Since the station count comes from your Orbit account (not hardcoded), the integration now routes the HT32A through the existing HT34A code path. Your three timers should discover and expose their two valves each instead of being skipped.

**Important:** I don't have an HT32A here, so actuation is **unverified** — same experimental caveat as the HT34A. Could you update and let me know:

1. In HACS, update **Orbit B-Hyve BLE** to v2.1.1 (redownload if it doesn't offer the update), then restart Home Assistant.
2. Confirm the three timers now appear and the `no device class for hardware='HT32A-0001'` error is gone.
3. Try opening/closing a valve. If it actuates and reports state correctly — we're done.

If a valve **doesn't** actuate: enable BLE snoop log on your phone, open the B-Hyve app, turn a station on then off, and send me the resulting bug report. That capture lets me confirm the exact protobuf/magic bytes the HT32A uses versus the HT34A.

[...]

PR #20, opened 2026-07-05
docs: correct run-progress field mapping (#16.#6.#5 remaining, #7 total)

exactmike:
Follow-up correction to #18.

On a hardware re-verification yesterday (timed live 180 s run, reading status a few times) the run-progress fields in `docs/ble-reliability-and-behavior.md` turned out to be **mislabeled**:

- **`#16.#6.#5` = seconds remaining** — it counts DOWN (measured `174 → 138 → 102`, once/sec), matching the vendor name `currentTimeRemainingSec`.
- **`#16.#6.#7` = total run time** — a constant (held at `180` the whole run), i.e. `totalRunTimeSec`, **not** remaining.
- **No Gen2↔XD layout flip.** Both a Gen2 (HT25G2 fw `0111`) and the XD (HT34A fw `0107`) behaved identically; the earlier "`#16.#7.#6` on Gen2" claim was wrong. `#16.#7` is empty or a small `{#6=1}` flag (`faultStatus`), not a countdown.

Because of that, the §1.5 "static remaining *firmware* quirk" was actually a decode reading the constant total field — not firmware. Corrected **§2.1**, **§1.5**, and the device-coverage table. The wall-clock drift-guard note stays as a reasonable safety net.

Docs-only; no code in this repo is affected. Thanks!

### DIFF
--- a/docs/ble-reliability-and-behavior.md
+++ b/docs/ble-reliability-and-behavior.md
@@ -70,11 +70,13 @@ per-poll re-subscription can **wedge** it. **Fix:** always send **`#57{#1=0}`** 
 done (e.g. in a `finally`). Separately, a dropped `#59` on a long-lived connection desyncs the RX
 counter; recover by **reconnecting** (a fresh handshake re-seeds IV + counter). **[HW-verified]**
 
-### 1.5 Some firmware reports a *static* `remaining`
-On a manual run, a valve may report the **same** `#16` remaining-seconds on **every** poll instead of
-counting down. A client that re-arms a wall-clock auto-close from each reading will push the end time
-forward forever and the entity appears to water indefinitely. **Fix:** drive the auto-close from the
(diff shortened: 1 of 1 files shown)


exactmike:
Added a second correction to this PR (§2.4 rain delay). A timed skew probe (set `#1=60min` but `#3=deviceClock+120s`, poll every ~30s) on both a Gen2 (fw`0111`) and the XD (fw`0107`) showed: run-state (`#16.#1`) flips `3→1` exactly when the clock crosses `#3` (ignoring `#1`), and `#16.#13` is **not** cleared on expiry — it lingers stale as `{#1:60,#3:<past>,#4:1}`. So "delay active" must be read from `#16.#1==3` (or a wall-clock check vs `#3`), not from the block's presence or `#1>0`. This supersedes the earlier `minutes>0` / "device omits the block" note in that section.

Issue #23, opened 2026-07-06
Design proposal: unify the protobuf-family devices + give them real device-read state (+ BLE reliability)


exactmike:
**Status:** design heads-up before a larger PR — I'd like a sanity check on the shape before I open the diff.

Hi @ljmerza — following on from the docs I contributed (#18) and the protocol reference you merged (#22), I have a substantial functional change staged and verified across my fleet (4× Gen2 HT25G2 fw0111 + an HT34A XD fw0107, over ESPHome proxies) and I'd rather agree the shape with you *first* than drop a big surprise PR. Here's the proposal; tell me if you'd prefer it whole or sliced.

## Motivation

Today the two protobuf devices — `ht25g2.py` (Gen2) and `ht34a.py` (XD, and now HT32A via #14) — are separate classes that already share the wire protocol. `ht34a.py`'s own docstring says it's "identical with HT25[G2]; only the inner plaintext and magic byte differ," and `ht25g2.py` already does:

```python
from .ht34a import _STOP_PB, _build_message, _build_start_pb
```

So the coupling exists; it's just split awkwardly across two files with the RX decode inline. Meanwhile the current status handling is still largely optimistic/ack-driven for Gen2. Now that the RX protocol is documented and hardware-verified (the #22 reference + #18/#20/#21), we can decode real device state instead.

## What I'd like to land

**1. Consolidate the protobuf family** into a single `BHyveProtobufDevice` (`devices/protobuf.py`); `ht25g2` / `ht34a` / `ht32a` become thin subclasses (frame magic byte + log label). **The mesh family (`ht25.py` / `ht25_fw0085.py`) is deliberately untouched** — this is scoped entirely to the protobuf pair.

**2. Real device-read status** in a shared `devices/status.py` decoder, per the merged #22 reference:
- run-state `#16.#1` (1 idle / 3 rain-delay / 4 running)
- `seconds_remaining` = `#16.#6.#5` (the #21 fix), total = `#16.#6.#7`
- active zone = `#16.#6.#4` (only the running zone renders open on the XD)
- `#30` `WateringStatus` enum (terminal/bare ⇒ idle; `{2,3,5,6,7}` ⇒ active) instead of idling on any `#30`
- battery mV via `#16.#14.#3`

[...]

PR #24, opened 2026-07-06
feat: unify the protobuf-family devices + real device-read state, reliability & capabilities

exactmike:
Implements the design proposal in #23 (you OK'd a single PR there).

Unifies the two protobuf-family device classes (Gen2 **HT25G2**, XD **HT34A/HT32A**) into one `BHyveProtobufDevice`, gives them real device-read state decoded from the HW-verified `#16` status block (per the protocol reference merged in #22), adds a systematic BLE-over-proxy reliability pass, and surfaces the rain-delay, flow, and battery capabilities. The mesh family (`ht25.py` / `ht25_fw0085.py`) is untouched except the shared ephemeral-session wrapper.

> **Scope note vs #23:** the issue framed this as foundation-first with *separate* follow-ups (rain delay / flow / battery). On extraction those turned out to be woven into the foundation at the protocol layer — flow/rain-delay are used as self-heal elicitors in the recovery path and decoded inline in the shared status parser — so splitting them would mean stripping and later re-weaving verified self-heal behavior. They come together here instead. Happy to slice differently if you'd prefer.

Hardware-verified across 4× Gen2 (HT25G2 fw0111) + the HT34A XD (fw0107) over ESPHome proxies.

## What's in it

**Consolidation**
- `devices/protobuf.py` (new): `BHyveProtobufDevice` — shared TX builders, actuation, `#15` `refresh_state`, self-heal recovery. `ht25g2.py` / `ht34a.py` become thin subclasses (frame magic + label). The two already shared the wire protocol (`ht25g2` imported `ht34a`'s builders); this finishes that coupling.
- `devices/status.py` (new): shared RX decoder (`extract_status`) — run-state (1 idle / 3 rain-delay / 4 running), battery mV (`#16.#14.#3`), `seconds_remaining` (`#16.#6.#5`, counts down), active zone (`#16.#6.#4`), `#30` `WateringStatus` enum, rain-delay (`#16.#13`), flow (`#59`).

**State-sync**
- `#15{}` is the canonical `refresh_state()`. The connect-time unsolicited push is reliable when idle but suppressed while the device is active (watering / rain-delay), so we elicit a full `#16` in any state — scheduled / app / physical-button runs become visible, not just HA commands.
- Wall-clock auto-close backstop (guarded to only ever move *earlier*); bare-`#30` stop-ack confirmed via a `#15` read-back; active zone so only the running station renders open on the XD.

**Reliability (BLE over ESPHome proxies)**
- Ephemeral connect-on-demand sessions under a per-device lock, disconnect in `finally`, so polls / commands don't share a session and desync the CTR stream.
- CTR-desync self-heal gated to a reply's first frame; GATT settle window; connection-health telemetry (`last_successful_poll` / `consecutive_timeouts`); single-level connect/handshake retry.

**Capabilities**
- **Rain delay** — `#17` TX + a "Rain delay" number (hours; 0 clears) + "Rain delay ends" sensor; active keyed on run-state, `#3` anchored to the device clock.
- **Flow (Gen2)** — `#57`/`#59` + a "Flow rate" gpm gauge (slope of the cumulative counter) + a "Check flow" button; a guaranteed `#57{0}` unsubscribe so a stream can't starve `#16`; configurable Flow-calibration option. The XD emits no `#59` (gated on `has_flow`).
- **Battery** — stop seeding from the cloud snapshot (BLE reports it live), restore the last live reading across restarts, "Watering ends" timestamp sensor.

## What this supersedes

On the **HT34A path this isn't purely additive** — it replaces the inline decoder you ported in #12. Your `ht34a.py` `_observe_plaintext`/`_parse_status` and the shared `status.py` decode the *same* `#16` fields with the same semantics (battery `#46→#3`, run-state `#16.#1`, remaining `#16.#6.#5` — your merged #21 fix); the shared decoder adds active-zone, rain-delay, `#30`, and flow, and brings the same real device-read state to **Gen2**, which today is optimistic-only (`start` ⇒ `is_watering = True`). So on the Gen2 side this is purely additive.

Two changes land on your hardware-verified XD path, flagged rather than buried in the diff:
1. **Rain-delay is now surfaced** — run-state `3` reads as rain-delay instead of just "not watering."
2. **Active zone comes from telemetry** — a program/app/button run (not initiated by HA) shows the actual running station (`#16.#6.#4`) instead of an optimistic guess.

## Backwards compatibility

- Mesh devices unchanged except the ephemeral-session wrapper.
- Existing entities keep their identities; this adds decoded state + a couple of diagnostic sensors.
- Gated by hardware/firmware/type via `resolve_device_class()`.
- `manifest.json` version left at `2.1.2` for you to bump on merge.

## Testing

Adds a `tests/` suite (new to the repo) that runs without Home Assistant or hardware — `conftest` registers a lightweight `orbit_bhyve` namespace package so the pure protocol/decode/dispatch logic is exercised directly. Covers frame/cipher round-trips, TX builders, RX status decode, device-class dispatch (notably Gen2 HT25G2 → protobuf, not the mesh class), ephemeral-session teardown, flow unsubscribe/cancel safety, and the single-level retry. **94 tests, all passing** (`python -m pytest`). Easy to drop from this PR if you'd rather land the suite separately.

### DIFF
--- a/custom_components/orbit_bhyve/__init__.py
+++ b/custom_components/orbit_bhyve/__init__.py
@@ -7,6 +7,7 @@ only at setup time and on user-triggered refresh; runtime is BLE-only.
 from __future__ import annotations
 
 import logging
+from datetime import timedelta
 from typing import Any
 
 import voluptuous as vol
@@ -22,11 +23,13 @@ from .const import (
     CONF_DEFAULT_DURATION,
     CONF_DEVICES,
     CONF_EMAIL,
+    CONF_FLOW_COUNTS_PER_GALLON,
     CONF_IDLE_DISCONNECT,
     CONF_PASSWORD,
     CONF_POLL_IDLE,
     CONF_POLL_WATERING,
     DEFAULT_DURATION,
+    DEFAULT_FLOW_COUNTS_PER_GALLON,
     DEFAULT_IDLE_DISCONNECT,
     DEFAULT_POLL_IDLE,
     DEFAULT_POLL_WATERING,
@@ -63,6 +66,9 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     idle_disconnect = opts.get(CONF_IDLE_DISCONNECT, DEFAULT_IDLE_DISCONNECT)
     poll_idle = opts.get(CONF_POLL_IDLE, DEFAULT_POLL_IDLE)
     poll_watering = opts.get(CONF_POLL_WATERING, DEFAULT_POLL_WATERING)
+    flow_counts_per_gallon = opts.get(
+        CONF_FLOW_COUNTS_PER_GALLON, DEFAULT_FLOW_COUNTS_PER_GALLON
+    )
 
     runtime = EntryRuntime()
     for record in devices:
@@ -72,7 +78,11 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         if (record.get("type") or "").lower() == "bridge":
             continue
         try:
-            device = build_device(hass, record, idle_disconnect_sec=idle_disconnect)
+            device = build_device(
+                hass, record,
+                idle_disconnect_sec=idle_disconnect,
+                flow_counts_per_gallon=flow_counts_per_gallon,
+            )
         except UnsupportedModel as err:
             _LOGGER.warning("%s: %s — skipping", record.get("name"), err)
             continue
@@ -88,6 +98,14 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
     entry.async_on_unload(entry.add_update_listener(_async_update_listener))
     _register_services(hass)
+
+    # Kick a prompt (non-blocking) first poll so an in-progress run / live state
+    # is read within one cycle instead of waiting a full idle interval. Matters
+    # after an HA restart mid-run: without this the valve shows idle until the
+    # next idle-cadence tick. async_request_refresh only *schedules* the poll, so
+    # this doesn't block setup on a slow/deep-sleeping device.
+    for coord in runtime.coordinators.values():
+        await coord.async_request_refresh()
     return True
 
 
@@ -104,8 +122,39 @@ async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
 
 async def _async_update_listener(hass: HomeAssistant, entry: ConfigEntry) -> None:
-    """Reload entry when options change."""
-    await hass.config_entries.async_reload(entry.entry_id)
+    """Apply changed options IN PLACE — no reload.
+
+    A reload would rebuild every device with a fresh idle DeviceState (wiping an
+    in-progress run's live state) and immediately disconnect+reconnect BLE. Orbit
+    valves allow only ONE BLE session, so reconnecting that fast — before the
+    device/proxy releases the old session — leaves comms broken until something
+    (a full HA restart) clears the stale session. So for our tuning options (flow
+    calibration, poll intervals, idle disconnect) we mutate the live coordinators
+    instead. Device-list/credential changes take a different path
+    (refresh_devices / reauth) that reloads explicitly."""
+    runtime: EntryRuntime | None = hass.data.get(DOMAIN, {}).get(entry.entry_id)
+    if runtime is None:
+        await hass.config_entries.async_reload(entry.entry_id)
+        return
+    opts = entry.options
+    idle_disconnect = opts.get(CONF_IDLE_DISCONNECT, DEFAULT_IDLE_DISCONNECT)
+    poll_idle = opts.get(CONF_POLL_IDLE, DEFAULT_POLL_IDLE)
+    poll_watering = opts.get(CONF_POLL_WATERING, DEFAULT_POLL_WATERING)
+    flow_counts_per_gallon = opts.get(
+        CONF_FLOW_COUNTS_PER_GALLON, DEFAULT_FLOW_COUNTS_PER_GALLON
+    )
+    for coord in runtime.coordinators.values():
+        coord.poll_idle = poll_idle
+        coord.poll_watering = poll_watering
+        coord.device.flow_counts_per_gallon = flow_counts_per_gallon
+        if coord.device.connection is not None:
+            coord.device.connection._idle_sec = idle_disconnect
+        # Re-apply the interval for the current state so a changed cadence takes
+        # effect now (the coordinator re-derives it each poll anyway).
+        coord.update_interval = timedelta(
+            seconds=poll_watering if coord.device.state.is_watering else poll_idle
+        )
+    _LOGGER.debug("%s: options applied in place (no reload)", entry.entry_id)
 
 
 def _register_services(hass: HomeAssistant) -> None:
--- a/custom_components/orbit_bhyve/binary_sensor.py
+++ b/custom_components/orbit_bhyve/binary_sensor.py
@@ -55,6 +55,11 @@ class _BHyveBinarySensorBase(CoordinatorEntity[BHyveDeviceCoordinator], BinarySe
 
 
 class BHyveConnectedBinarySensor(_BHyveBinarySensorBase):
+    """Connectivity: True when the last status poll REACHED the device. Under the
+    ephemeral connect-on-demand model the BLE link is torn down between polls, so
+    this reflects poll-reachability (coherent with the Consecutive timeouts sensor)
+    rather than a live socket."""
+
     _attr_device_class = BinarySensorDeviceClass.CONNECTIVITY
     _attr_entity_category = EntityCategory.DIAGNOSTIC
 
--- a/custom_components/orbit_bhyve/button.py
+++ b/custom_components/orbit_bhyve/button.py
@@ -1,14 +1,15 @@
 """Button platform — per-device sync button.
 
 One ButtonEntity per non-hub device. Pressing it forces a fresh BLE
-connect + 8-step init; the info-ack notification that arrives during
-init carries battery_mV (parsed by devices.base.BHyveBleDeviceBase
-._observe_plaintext), and the coordinator refresh after the press
-pushes the new value into HA.
+connect + 8-step init, then requests a coordinator refresh — which for
+protobuf devices issues a solicited #15 status read (real run-state,
+battery, rain-delay, seconds-remaining), the reliable way to pull live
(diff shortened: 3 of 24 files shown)


ljmerza:
Pulled this onto my box and ran it live against my three mesh **HT25-0000** valves (Deck fw0085; Hill + Corner fw0041 — all the mesh path). Setup loads clean, all three coordinators poll `success: True`, entities and battery/diagnostics look right. But **every valve start/stop throws and the valve never actuates.** Sharing the root cause since it's on the mesh slice you flagged as the widest blast radius.

**Symptom** (open or close, both fail identically):

```
CTR desync detected (bad header d747c105) — scheduling disconnect to self-heal
...
File "custom_components/orbit_bhyve/devices/ht25.py", line 185, in _post_handshake
    await conn._write_locked(plaintext)
File "custom_components/orbit_bhyve/connection.py", line 302, in encrypt
File "custom_components/orbit_bhyve/connection.py", line 298, in _aes_xor
File "custom_components/orbit_bhyve/connection.py", line 292, in _aes_keystream
    out.extend(encryptor.update(self._iv + struct.pack("<I", c)))
TypeError: unsupported operand type(s) for +: 'NoneType' and 'bytes'
```

**Root cause — the new desync self-heal misfires on mesh replies, then races the handshake:**

1. The new `_on_notify` self-heal treats any first-frame reply that doesn't start with the protobuf inner-message header `aa 77 5a 0f` as a CTR desync:
   ```python
   if len(pt) >= 4 and pt[:4] != b"\xaa\x77\x5a\x0f":   # connection.py
       ...
       self.hass.add_job(self.disconnect)
   ```
2. Mesh HT25 replies don't use that header — they're `<mesh_addr:2><type:1>[seq][routing][payload]`, e.g. `d747 c1 05 40 …`, exactly as `base.py::_observe_plaintext` documents. So the very first `bind`-step reply in `_post_handshake` is misclassified as a desync and schedules `disconnect()`.
3. `disconnect()` nulls `self._iv` / `self._client`. Because it's scheduled via `hass.add_job`, it interleaves with the still-running 8-step `_post_handshake` loop across its `await asyncio.sleep(INIT_INTER_STEP_SEC)`. The next `_write_locked → encrypt → _aes_keystream` then runs with `self._iv = None` → the `TypeError` above (also surfaces at `_write_locked`'s `wait_for` when `_client` is gone first).
4. `send_actuation` re-runs the post-handshake hook on every command, so this reproduces on 100% of open/close attempts. (No valve is left stuck — the command just hard-fails.)

**Suggested fix:** make the reply-header check device-class-aware, the same way `frame_magic` / `trailer_const` are already per-class knobs threaded into `BHyveBleConnection`. e.g. add a `reply_header: bytes | None` — `b"\xaa\x77\x5a\x0f"` for the protobuf classes, `None` for the mesh HT25 classes (skip the header-based desync check). Also worth a guard so `encrypt` / `_write_locked` bail cleanly if `self._iv is None`, so a legitimately-scheduled mid-handshake disconnect can't crash the init sequence.

One secondary observation: after that fix, mesh would have no working desync self-heal at all — the `except`-based "decrypt failed" branch won't fire on a real mesh CTR desync, since `decrypt()` only raises on a bad `frame_magic`, not on garbage plaintext. If mesh self-heal is wanted, validating against the mesh frame shape (`pt[4] == 0x40 and pt[2] & 0x40`, per `_observe_plaintext`) would be a truer signal than the protobuf header.

Happy to test any patch on the mesh hardware. The protobuf/Gen2/XD paths I can't exercise (no hardware), so those still rest on your fleet testing.

anahnymous:
Confirmed working on HT34A-0001 Firmware: 0107.  Walked through all 4 zones.  Turning on a new zone when another is active disabled the previous zone.  This is the only model I have to test.

exactmike:
Thanks for the detailed root cause — it was exactly right, and it pointed straight at the fix. Pushed as `f856bcc`.

**What changed**

- Added a per-class `reply_header: bytes | None` knob, threaded into `BHyveBleConnection` the same way `frame_magic`/`trailer_const` are. The protobuf classes set the inner-message header `b"\xaa\x77\x5a\x0f"`; the mesh HT25 classes leave it `None`, which skips the header-based desync check entirely. So the mesh bind-step reply is no longer misclassified.
- Guarded `encrypt()` to raise a typed `BleHandshakeError` (which the open-retry already catches) instead of a `TypeError` from `_aes_keystream` if a scheduled disconnect clears `_iv` mid-handshake — defense-in-depth for exactly the race you traced.

**On your secondary point (mesh has no self-heal now):** agreed, and I left it that way deliberately rather than add the frame-shape heuristic. The mesh `start_watering`/`stop_watering` disconnect in a `finally`, so every mesh actuation is a fresh cold handshake — there's no long-lived pooled session accumulating CTR desync (that pooled-streaming case is what motivated the self-heal on the protobuf/flow path). So mesh is structurally much less exposed. Happy to add a `pt[4]==0x40 and pt[2]&0x40` mesh-shape check if you'd rather have belt-and-suspenders, but I didn't want to ship an unverified heuristic into your one working path.

**Verified on our end (protobuf only — we have no mesh HT25-0000 hardware):** clean start/stop on a Gen2 HT25G2 and the HT34A XD. Notably the self-heal still works where it should — on one close, the first frame after a fresh handshake came back a genuine desync (`bad header`), the self-heal fired, reconnected, re-handshook, and `STOP confirmed idle`. So the header check is intact for protobuf; it's only suppressed for mesh.

**While I was in there** I did a full pass over the mesh path against the protocol spec and found no other cross-family gotchas — the frame builders, the 8-step init, start/stop, seq IDs, and the battery decode all line up. One thing only you can settle, though: the fw0085 rebind uses `_rebind_sid_delta = 3` (vs `2` for fw0041), and that value has never been hardware-verified here — it's carried over from the legacy fork. So **if fw0085 still doesn't actuate after this fix, the next suspect is the rebind delta — try flipping it to 2.** fw0041 should be unaffected either way.

Would appreciate a retest on the mesh valves when you get a chance.

exactmike:
@anahnymous thank you — a full 4-zone walk-through on the HT34A-0001 (fw0107) is exactly the independent verification we needed, since it's the same protobuf family as the Gen2 valves we tested here.

On the behavior you noticed — **turning on a new zone disabling the previously-active one is expected HT34A hardware behavior, not an integration bug.** The HT34A is a single-manifold multi-station timer: it runs one station at a time, so starting a new zone stops the prior. Three things line up with that:

- The status protocol reports a single active station (`wateringStatus.currentStationId`), not a per-zone bitmap.
- Stored programs step through their zones **sequentially** for the same reason (visible in the XD program captures).
- Our `start_watering` sends **only** the new station's start command — it never issues a stop to the other zones. The device does the switching itself.

And the integration models it correctly: on a start we set the decoded active station, and each valve entity renders open only when it *is* the running station — so the previously-running zone flips to closed on its own, no lingering "both zones open" state. So this is working as intended; nothing to change.

Really appreciate the testing.

ljmerza:
Hardware-tested this on my mesh fleet (HT25-0000: Deck `fw0085`, Hill/Corner `fw0041`) over an ESPHome BLE proxy.

## ✅ `f856bcc` fixes the mesh crash
The device-class-aware self-heal resolves the CTR-desync crash — mesh start/stop no longer trips the `reply_header` check during the d7-47 bind, and `encrypt()` no longer throws `TypeError` on a cleared session. Confirmed on hardware (the part you couldn't verify on your end). Integration loads clean, all 3 coordinators poll `success: True`, 33 entities with the protobuf-only ones correctly gated out.

## ❌ New issue on **stop**: stale pooled session raises `BleakError: Not connected` (uncaught → service-call error)
Separate from the crash above. Stopping a zone surfaced this to the HA frontend as an `Unexpected exception`:

```
aioesphomeapi.core.BluetoothGATTAPIError: Bluetooth GATT Error
    address=44:67:55:52:99:01 handle=11 error=-1 description=Not connected
  → bleak.exc.BleakError: ... Not connected
```

Call chain (all in-component):
```
valve.py async_close_valve
  → ht25.py stop_watering → _send_command
    → connection.py send_actuation        # pooled session looked alive → re-bind in place
      → ht25.py _post_handshake
        → connection.py _write_locked → bleak write_gatt_char → GATT "Not connected"
```

**Likely cause:** `send_actuation` takes the *"pooled connection — refresh the bind in place"* branch when `is_connected and _handshaken` is `True`, but over the ESPHome proxy the link can be **dead at the GATT layer while HA still thinks it's up**. The in-place `_write_locked` then fails `Not connected`, and nothing catches it — the mesh `ht25.stop_watering`/`_send_command` path has `try/finally: disconnect` but no `except`, so the `BleakError` propagates to the `close_valve` service call.

**It's transient / recovers:** a subsequent attempt did a fresh `connecting (attempt 1/3)` → full 8-step init → `STOP ack — cleared off-timer`, so the valve *did* close — but the first attempt threw a user-visible error.

**Suggested direction:** have `send_actuation` catch a `BleakError`/timeout on the pooled-refresh write and fall back to `disconnect()` + fresh `ensure_connected()` + retry (the protobuf class already self-retries on a fresh session; the mesh path relies on `send_actuation` and has no equivalent catch). Happy to test a patch on my hardware.

<sub>🤖 hardware test + log capture via Claude Code</sub>

ljmerza:
Hardware-tested the fix (`f856bcc`, device-class-aware desync) on my mesh fleet — **it works.** The original `TypeError: NoneType + bytes` crash is gone and mesh actuation is functional end to end:

- **Deck (fw0085, `BHyveHT25Fw0085Device`)** — turns ON and OFF cleanly.
- **Corner (fw0041, `BHyveHT25Device`)** — turns ON and OFF cleanly (flawless run: START ack → off-timer armed → STOP ack cleared it).
- All 8 mesh init steps (bind/status/info/subsystem/magic1/magic2/heartbeat/rebind) now parse with **zero false desyncs** — the `d747c105…` bind replies that previously tripped the self-heal are handled correctly.
- Battery now reads a live value off the device (Deck 31% @ 2588 mV, Corner 37% @ 2623 mV) instead of crashing the handshake.

So the mesh regression is resolved and this is good to merge from the mesh side. 👍 (The protobuf/Gen2/XD paths I still can't exercise — those remain on your fleet testing.)

Two **non-blocking** mesh follow-ups surfaced during testing — filed as #25:

1. **Back-to-back actuation** (open then close within ~30 ms) throws `BluetoothGATTAPIError: Not connected` — `finally: disconnect()` isn't lock-guarded and races the next command's pooled reuse. Self-recovers on retry; only bites truly pipelined commands, not normal sequential use.
2. **Mesh state only refreshes on actuation** — after an HA restart, an un-actuated mesh valve's battery/state sit `unknown` (mesh `refresh_state` doesn't connect + the cloud battery seed was dropped), and the **Sync button is a no-op on mesh** for the same reason.

Details, root cause, and fix caveats are in #25.

Issue #25, opened 2026-07-06
Mesh HT25 follow-ups after #24: back-to-back actuation GATT race + mesh state only refreshes on actuation (unknown after restart, Sync no-op)


ljmerza:
## Context

While hardware-testing #24 on 3× mesh **HT25-0000** valves (Deck fw0085 = `BHyveHT25Fw0085Device`; Hill + Corner fw0041 = `BHyveHT25Device`), the headline fix from #24 (device-class-aware CTR-desync self-heal, `f856bcc`) is **confirmed working** — both firmware classes actuate ON and OFF cleanly, all 8 mesh init steps parse with zero false desyncs, and battery now reads a live value. 

Two mesh-path issues surfaced during that testing. Neither is a merge-blocker for #24 — they're follow-ups. Filing together since both stem from mesh BLE connection handling; happy to split into two issues if preferred.

Environment: HA on Python 3.14, BLE via ESPHome Bluetooth proxy.

---

## Problem 1 — Back-to-back actuation throws `BluetoothGATTAPIError: Not connected`

**Repro:** issue an open then a close on the same mesh valve within ~30 ms (e.g. a dashboard double-fire or an automation that opens then closes).

**Observed:** the START succeeds and the valve turns on; the STOP fired 29 ms later fails:
```
START ack — armed off-timer for 7200s          # valve ON
... 29ms later ...
aioesphomeapi.core.BluetoothGATTAPIError: Bluetooth GATT Error
    address=44:67:55:52:99:01 handle=11 error=-1 description=Not connected
```
A subsequent (cold) retry succeeds, so it self-heals — but the failure surfaces to the UI as an "Unexpected exception."

**Root cause:** `start_watering`/`stop_watering` wrap the command in `finally: await connection.disconnect()` (`devices/ht25.py:207` / `:220`), but `disconnect()` (`connection.py:252`) does **not** hold `self._lock`. So command #1's `finally: disconnect()` runs concurrently with command #2's `send_actuation`, whose pooled-reuse fast-path (`connection.py:387-390`) saw `is_connected == True` and reuses a link that #1's disconnect then tears down → the next GATT write hits "Not connected."

**Also:** `_send_command` catches `TimeoutError` but not `BleakError`/`BluetoothGATTAPIError` (`devices/ht25.py:232`), so the GATT error propagates raw to the service call instead of being handled.

[...]

PR #26, opened 2026-07-06
fix(connection): reopen a fresh session when a pooled actuation bind fails (mesh STOP over proxy)

ljmerza:
## What

Fixes the stale-pooled-session actuation failure I reported on #24 (https://github.com/ljmerza/orbit-bhyve-ble/pull/24#issuecomment-4894290334), found while hardware-testing #24 on live mesh HT25 valves over an ESPHome proxy.

> **Stacked on #24.** Base is `pr24-base`, a snapshot of #24's head (`f856bcc`), so this PR's diff is **just the one fix commit** — meant to be folded into #24 (cherry-pick `d398267`) or merged after it. `pr24-base` is a disposable branch. The bug pre-exists on `main` too (same uncaught pooled-refresh), so this can also target `main` if you'd rather land it independently.

## The bug

`send_actuation` refreshes the per-class bind **in place** on a pooled connection:

```python
if self.is_connected and self._handshaken:
    if self._post_handshake_hook is not None:
        await self._post_handshake_hook(self)   # ← writes over a possibly-dead link
```

Over an ESPHome proxy the pooled link can be **dead at the GATT layer while `is_connected` still reads `True`**, so those refresh writes fail:

```
aioesphomeapi BluetoothGATTAPIError: ... handle=11 error=-1 description=Not connected
  → bleak.exc.BleakError: ... Not connected
```

That `BleakError` propagated **uncaught** — the mesh `ht25._send_command` only tolerates a write-response `TimeoutError`, not a `BleakError` — up through `stop_watering` → `valve.async_close_valve`, surfacing in HA as an `Unexpected exception`. A manual retry (fresh connect → full init → `STOP ack`) then succeeded, confirming the pooled session was simply stale.

## The fix

Catch `BleakError` / `OSError` / `TimeoutError` from the pooled bind-refresh in `send_actuation`, drop the stale session, and reopen a fresh one (`ensure_connected` re-runs the hook) before writing the command — mirroring the recovery that already worked by hand. It's a shared-connection-layer fix; the mesh path (which relies on `send_actuation`) is the one that broke. Protobuf classes go through `send()` + their own disconnect/retry and are unaffected.

## Testing

- `python -m pytest` → **97 passed** (adds one regression test: simulate a `BleakError` on the pooled bind-refresh, assert `send_actuation` drops the session, reopens fresh, and still writes the command instead of raising).
- Hardware validation on mesh HT25 is pending — the failure is intermittent (needs a genuinely dead pooled link), so I'm relying on the unit test plus the matching hardware trace from the report. Happy to deploy + soak it on my fleet.

<sub>🤖 Generated with Claude Code</sub>

### DIFF
--- a/custom_components/orbit_bhyve/connection.py
+++ b/custom_components/orbit_bhyve/connection.py
@@ -405,8 +405,22 @@ class BHyveBleConnection:
         async with self._lock:
             if self.is_connected and self._handshaken:
                 # Pooled connection — refresh the (possibly stale) bind in place.
-                if self._post_handshake_hook is not None:
-                    await self._post_handshake_hook(self)
+                # But over an ESPHome proxy the pooled link can already be dead at
+                # the GATT layer while is_connected still reads True, so the refresh
+                # writes fail with BleakError ("Not connected"). Don't let that
+                # surface as a failed actuation: drop the stale session and reopen
+                # a fresh one (ensure_connected re-runs the hook). HW-observed on a
+                # mesh HT25 STOP over a proxy — the very next fresh connect landed.
+                try:
+                    if self._post_handshake_hook is not None:
+                        await self._post_handshake_hook(self)
+                except (BleakError, OSError, asyncio.TimeoutError) as err:
+                    _LOGGER.debug(
+                        "%s: pooled bind refresh failed (%s) — reopening a fresh session",
+                        self.mac, err,
+                    )
+                    await self.disconnect()
+                    await self.ensure_connected()
             else:
                 # Cold — ensure_connected() retries the open and runs the hook.
                 await self.ensure_connected()
--- a/tests/test_devices.py
+++ b/tests/test_devices.py
@@ -733,3 +733,61 @@ def test_ensure_connected_does_not_multiply_retries(monkeypatch):
         asyncio.run(conn.ensure_connected())
 
     assert calls["connect"] == OPEN_MAX_ATTEMPTS  # 3, never 9
+
+
+# --- stale pooled session self-heals on actuation (connection.py) ----------
+
+def test_send_actuation_reopens_when_pooled_bind_refresh_fails():
+    # Regression (ljmerza PR #24 hardware feedback): over an ESPHome proxy a
+    # pooled session can be dead at the GATT layer while is_connected still reads
+    # True. send_actuation's in-place bind refresh then fails with BleakError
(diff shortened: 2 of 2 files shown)


exactmike:
This recovery approach holds up — I'd independently built the same catch-and-reopen while working the #24 follow-up, so 👍 on the direction. One note on the root cause, though, tied to your own #25 write-up.

In #25 you diagnosed P1's root cause as the **concurrent-teardown race**: `start`/`stop_watering`'s `finally: await connection.disconnect()` runs **outside** `self._lock`, so command #1's teardown races command #2's `send_actuation` pooled fast-path (it reused the link that #1 then tears down). This #26 *recovers* from that (the victim reopens fresh), but leaves the race itself in place — `disconnect()` is still lock-free — so a rapid open→close still eats a wasted reconnect cycle, and the catch also trips on the benign in-flight case rather than the genuinely-dead-link case it's meant for.

If you want to close the race at the source rather than just self-heal it, the minimal tweak is to make the public `disconnect()` acquire the lock while keeping a lock-free `_teardown()` for the callers that already hold it — which is exactly the `_open()` self-deadlock you flagged in #25:

```python
async def disconnect(self) -> None:
    """Public teardown — acquires self._lock so it can't race an in-flight
    send/actuation."""
    async with self._lock:
        await self._teardown()

async def _teardown(self) -> None:
    """Lock-free body (the current disconnect()). Caller holds self._lock."""
    ...
```

Then the in-lock callers (`_open()`'s retry, `_idle_close`, and this new fallback) call `_teardown()`; the external ones (the actuation `finally`, the `_on_notify` self-heal `hass.add_job(self.disconnect)`) call the locked `disconnect()`. Concurrent actuations then serialize instead of racing.

It's **complementary to #26, not a replacement** — your catch still handles a genuinely dead pooled link (proxy GATT drop with `is_connected` still True), the lock just removes the self-inflicted race so it stops firing on the benign case. Two small riders if you fold it in: wrap the pooled **write** too (not only the bind-refresh hook — the link can die between the refresh and the write), and have mesh `_send_command` catch `BleakError` (not just `TimeoutError`) as a backstop.

I've got this implemented + unit-tested (disconnect lock-serialization, the fallback for both hook and write, and the `_send_command` catch) but only protobuf hardware here — no HT25 mesh to soak it on, and this is your path. Happy to send it as a follow-up PR on top of #26, or just leave the sketch for you to fold in — your call.

PR #27, opened 2026-07-06
fix(button): Sync button forces a BLE connect on mesh devices

ljmerza:
## What

Restores the **Sync button's on-demand refresh on mesh devices** — a regression from #24. Found while hardware-testing #24 on live mesh HT25 valves.

Independent of #26 (that's the `connection.py` stale-pool actuation self-heal); this touches `button.py` / `base.py` / `ht25.py`. Both target `main`.

## The bug

#24 changed the Sync button to just call `async_request_refresh()`, dropping the old forced connect:

```python
# before #24:
await conn.disconnect()
await conn.ensure_connected()   # ← forced connect; connect-time init refreshed battery
await self.coordinator.async_request_refresh()
```

On **protobuf** devices (Gen2/XD) that's fine — their `refresh_state()` connects for a `#15` status read. But the **mesh** `refresh_state()` is deliberately passive (*"update is_connected … without opening one"*), so on mesh the Sync button became a **no-op**: pressing it never connects to the valve, so battery/state are never refreshed. Confirmed on hardware — the press logs `Finished fetching … in 0.000 seconds` with no BLE connect. (The button's own docstring still promises the connect.)

## The fix

Add an `async_manual_sync()` hook, run by the button before the coordinator refresh:
- **base**: no-op (protobuf's `refresh_state` already connects — no double connect).
- **mesh (`ht25`)**: force a fresh connect, so the connect-time 8-step init's `status` (seq `0x02`) and `info` (seq `0x03`) replies refresh watering-state + battery via `_observe_plaintext`, then disconnect (ephemeral).

## Testing

`python -m pytest` → **98 passed**. Adds two regression tests: mesh `async_manual_sync` forces exactly one connect + tears it down; the base hook is a no-op for protobuf (no second connect).

Hardware validation on mesh HT25 to follow.

<sub>🤖 Generated with Claude Code</sub>

### DIFF
--- a/custom_components/orbit_bhyve/button.py
+++ b/custom_components/orbit_bhyve/button.py
@@ -74,6 +74,10 @@ class BHyveSyncButton(CoordinatorEntity[BHyveDeviceCoordinator], ButtonEntity):
         if device.connection is None:
             return
         _LOGGER.info("%s: sync requested via button", device.mac)
+        # Force any device-class-specific on-demand refresh first (mesh forces a
+        # BLE connect so the connect-time init pulls fresh battery/state; protobuf
+        # is a no-op here since its refresh_state connects for a #15 read below).
+        await device.async_manual_sync()
         await self.coordinator.async_request_refresh()
 
 
--- a/custom_components/orbit_bhyve/devices/base.py
+++ b/custom_components/orbit_bhyve/devices/base.py
@@ -138,6 +138,13 @@ class BHyveBleDeviceBase(abc.ABC):
     async def async_setup(self) -> None:
         """Hook for device classes that want pre-warming. Default: no-op."""
 
+    async def async_manual_sync(self) -> None:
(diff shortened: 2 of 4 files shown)


exactmike:
Approach looks right to me — a `base` no-op + a mesh override that forces an ephemeral connect is the clean split, since the protobuf `refresh_state()` already connects for its `#15` read and would otherwise double-connect.

Worth calling out for anyone reading: this also fixes the **`unknown`-after-restart** half of #25 P2, not just the no-op button — the connect-time 8-step init's `status` (seq `0x02`) + `info` (seq `0x03`) replies backfill watering-state and battery via `_observe_plaintext`, so a single Sync press repopulates a mesh valve that hasn't actuated since restart. Nice.

No mesh hardware on my end to verify, but the code path matches what I'd expect. 👍

Issue #19, opened 2026-07-05
Offering: BLE transport-reliability + hardware-verified behavior notes (Gen2 + XD)


exactmike:
## Offering: BLE transport-reliability + hardware-verified behavior notes (Gen2 + XD)

We've been controlling the Orbit B-Hyve **Gen2 (HT25G2)** and **XD (HT34A)** valves locally over BLE
(via ESPHome proxies) and verifying behavior on hardware we own — 4× Gen2 fw `0111` + a 4-station XD
HT34A fw `0107`. The base protocol is already well documented elsewhere (see below); what we'd like
to contribute is the **operational layer on top** — the reliability behavior and fine-grained,
hardware-verified status semantics — as a small, complementary doc.

I've opened a **docs-only PR** (#18) adding `docs/ble-reliability-and-behavior.md`.

> **Update (2026-07-05):** #18 is merged. A hardware re-verification then corrected the
> run-progress field mapping (see the two struck-through items below) — fix in #20.

### What's in it (the parts not covered by the existing protocol docs)
**Transport reliability over BLE / proxies:**
- one BLE session at a time → ephemeral connect-per-op beats a pooled session;
- solicited vs. unsolicited RX (a `#15{}` elicitor is needed to read state while watering/rain-delayed);
- **duplicate-RX-frame CTR desync** (a proxy re-delivers a frame; the whole CTR stream then fails CRC) + the byte-identical dedup fix;
- a **persistent `#57` flow subscription that suppresses `#16`** and can wedge the valve (+ the `#57{#1=0}` fix), and `#59`-desync recovery via reconnect;
- a wall-clock auto-close safety net; and a CTR-desync self-heal. *(~~firmware reports a static `remaining` every poll~~ — corrected in #20: that was our own decode artifact reading the constant total field, not firmware.)*

**Hardware-verified status-decode semantics:**
- run progress `#16.#6`: **`#5` = remaining (counts down), `#7` = total** — same on Gen2 + XD, **no layout flip** *(~~Gen2↔XD flip `#16.#7.#6` vs `#16.#6.#7`~~ — corrected in #20)*;
- **active zone** `#16.#2.#2.#3.#1` (the only running-station report on the XD);
- STOP = **bare `#30` ack** → confirm via a follow-up `#15{}`;
- rain delay honors an **absolute `#3` expiry** (anchor to the device clock `#7`) + the bare-clear `#4`-absent gotcha;
[...]

PR #31, opened 2026-07-08
Watering programs + protobuf-family capability round-up

exactmike:
# Watering programs + protobuf-family capability round-up

Builds on #24 (protobuf-family foundation, now merged) and targets `main` at
v2.2.0. Adds local BLE **watering-program** control for the protobuf family
(HT25G2 Gen2 + HT34A XD), plus the supporting protocol/transport work. All
additive and gated to the protobuf family via `resolve_device_class()`; the
mesh path and existing behavior are untouched.

Hardware-verified end-to-end on 4× Gen2 (HT25G2 fw0111) + the XD (HT34A fw0107)
over ESPHome BLE proxies. 136 tests, no HA/hardware required to run them.

## What's in it

**Watering programs (the headline)**
- Read/write non-Smart program slots (A–F) over BLE: watering days
  (weekdays / even / odd / every-N-days), start time(s), per-zone run
  durations. Store, replace, enable/disable, delete.
- Services: `set_program`, `delete_program`, `get_program`
  (`SupportsResponse.ONLY`, returns the stored slot as structured data).
- Entities: **Automatic-watering** switch (controller auto/off),
  **Program A–D** enable switches (a slot with no stored program reads
  `unavailable`), **Program A–D** summary sensors (schedule in attributes),
  and a **Next run** timestamp sensor (device-computed next start).
- Slots are read back from the device's `#10` sync dump (a multi-frame reply),
  so programs created in the B-Hyve app show up too. There is no schedule
  *read* opcode in the protocol; `#10` is the only path and it returns all
  slots.
- Enabling a program uses the hardware-verified **3-write sequence** (store
  `#19` → enable `#20` bitmask → `#14 timerMode autoMode` with the required
  empty `manualModeParams` marker, re-sent), which is what the app does and
  the only sequence that makes a stored program actually compute a next start.

**Identify button** — `#47` LED-locate across the protobuf family (flashes the
Gen2 LED; no-op on the XD, which ignores it).

**Supporting protocol/transport work** (needed by programs, but general):
- `connection.send_stream()` — multi-frame RX reassembly: concatenates a
  reply's per-frame plaintext so a long `#10` dump that CTR-streams across
  frames can be reassembled. Single-frame paths (`#16`/`#30`/`#59`) unchanged.
- Program/mode writes fail gracefully on a marginal link. A CTR-desync
  self-heal can tear the pooled session down mid-write-sequence (the next write
  then raises `BleakError("Not connected")` over an ESPHome proxy); the write
  methods now catch that and return `False` (→ "not confirmed" + a coordinator
  refresh that re-reads real state) instead of surfacing a raw exception to the
  service/switch call. Mirrors the read path's best-effort behavior.

## Design choices & rationale

Two threads run through these changes: **keep leaning away from the cloud and
the native app**, and **add capability without over-committing HA surface**
before the shape is proven on hardware.

- **Time sync on the device, from HA.** Schedules, rain-delay expiry, and
  on-device auto-close all key off the device's own clock. Rather than depend
  on the app to keep that clock honest, the coordinator poll now syncs it to
  HA-now via `#75 setEpochTime` — which returns the same `#16` status block, so
  it costs nothing extra and folds another job the app used to own into the
  integration.

- **Programs: make them visible and workable, but keep the HA surface small
  for now.** This deliberately does *not* build out a full HA entity model for
  every program attribute (per-start-time entities, per-zone duration numbers,
  day-mode selectors, etc.). Instead it exposes programs through a compact set
  — enable switches, a read-only schedule summary sensor, a next-run sensor,
  and get/set/delete services — so a user can see what's stored (including
  app-created programs), toggle slots, and script changes today, without
  locking in a large entity schema before the ergonomics are proven. The full
  encode/decode is already in place under the hood; the entity surface can grow
  later without protocol rework.

## Behavioral change to call out (not purely additive)

One existing behavior changes here — flagging it explicitly for review
(rationale in *Design choices* above):

- **Coordinator poll elicitor: `#75 setEpochTime` instead of a bare `#15{}`.**
  Returns the same `#16` status block, and additionally syncs the device
  clock to HA-now. Content-identical status reply verified on hardware.

## Compatibility / safety

- Programs, switches, program sensors, Next-run, and Identify are created only
  for protobuf-family devices; mesh HT25 and the hub are unaffected.
- No new cloud dependency — everything is BLE after setup.
- Firmware caveat unchanged: RE'd against fw0111 / fw0107.

## Not included (stays in our fork)

The CLI (`scripts/bhyve.py`) and RE/exploration tooling that we develop and
hardware-verify against are intentionally left out of the integration, per the
repo's "diagnostic code stays out of the component" guidance. The shipping
program encoders/decoders live in `devices/protobuf.py` + `devices/status.py`.

### DIFF
--- a/custom_components/orbit_bhyve/devices/protobuf.py
+++ b/custom_components/orbit_bhyve/devices/protobuf.py
@@ -17,6 +17,7 @@ re-declared, so there is a single source for both directions.
 from __future__ import annotations
 
 import asyncio
+import functools
 import logging
 import struct
 import time
@@ -32,6 +33,31 @@ from .base import (
 from .status import MSG_HEADER, _crc16_ccitt, apply_status_plaintext, parse_sync_dump
 
 _LOGGER = logging.getLogger(__name__)
+
+
+def _ble_write_guard(method):
+    """Make a bool-returning BLE write method fail gracefully on a marginal link.
+
+    A CTR-desync self-heal tears the session down mid-sequence (its disconnect is
+    scheduled on the loop), so a following write raises BleakError("Not connected")
+    — over an ESPHome proxy this is a routine, transient event. Catch it (and a
+    proxy write-response timeout) and return False instead of surfacing a raw
+    exception to the service/switch call: the coordinator refresh that follows
+    re-reads the device's actual state, and the user/automation can retry. Mirrors
+    the read path's best-effort behavior (get_programs falls back to last-known).
+    The method's own try/finally has already run its disconnect by the time we
+    catch here."""
+    @functools.wraps(method)
+    async def _wrapped(self, *args, **kwargs):
+        try:
+            return await method(self, *args, **kwargs)
+        except (BleakError, asyncio.TimeoutError) as err:
+            _LOGGER.warning(
+                "%s: %s %s failed on a BLE error (%s) — marginal link, not confirmed",
+                self.mac, self.log_label, method.__name__, err,
+            )
+            return False
+    return _wrapped
 
 
 def _pb_varint(val: int) -> bytes:
@@ -654,6 +680,7 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
             finally:
                 await self.connection.disconnect()
 
+    @_ble_write_guard
     async def set_program(self, spec: ProgramSpec) -> bool:
         """Store a program (#19) and, if spec.enabled, drive the 3-write run
         handshake (store -> enable #20 -> autoMode #14{1} -> re-send store+enable).
@@ -707,6 +734,7 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
             finally:
                 await self.connection.disconnect()
 
+    @_ble_write_guard
     async def delete_program(self, slot: int) -> bool:
         """Clear a slot: drop its enable bit, write the #19 NotSet body, keep the
         controller in autoMode. Returns True once the read-back shows it empty."""
@@ -728,6 +756,7 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
             finally:
                 await self.connection.disconnect()
 
+    @_ble_write_guard
     async def set_program_enabled(self, slot: int, on: bool) -> bool:
         """Toggle a stored program's enable bit (#20 bitmask) and keep the
         controller in autoMode. On enable, re-send the bitmask while in autoMode so
@@ -766,6 +795,7 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
         next-start is reported."""
         return await self.set_program_enabled(slot, True)
 
+    @_ble_write_guard
     async def set_controller_mode(self, on: bool) -> bool:
         """Device-global controller mode via #14 timerMode: on -> autoMode(1)
         ('Enable Watering', scheduled programs run), off -> offMode(0) (automatic
@@ -783,6 +813,7 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
             finally:
                 await self.connection.disconnect()
 
+    @_ble_write_guard
     async def identify(self, seconds: int = 6) -> bool:
         """Flash the device's LED to locate it (#47 identifyDevice). The flash
         LATCHES on Gen2 (identifyTimeSec isn't a reliable auto-off), so we start it,
--- a/custom_components/orbit_bhyve/devices/protobuf.py
+++ b/custom_components/orbit_bhyve/devices/protobuf.py
@@ -664,7 +664,7 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
         stream = await self.connection.send_stream(
             _build_message(_build_sync_request_pb()), drain_ms=6000
         )
-        programs, mask, _status = parse_sync_dump(stream)
+        programs, mask, status = parse_sync_dump(stream)
         if programs:
             if mask is None:
                 # Mask frame dropped on this burst — keep each slot's known enable
@@ -674,6 +674,15 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
                     if sch.enabled is None and prev is not None:
                         sch.enabled = prev.enabled
             self.state.programs = programs
+        elif mask is not None or status is not None:
+            # ZERO #19 bodies but the burst DID decode a coherent frame (the #20 enable
+            # mask and/or the #16 status): the device answered fully and stores no
+            # programs — e.g. every slot deleted via the app on hardware that omits
+            # deleted slots rather than echoing NotSet bodies. Clear last-known so a
+            # stale schedule doesn't linger forever. A truncated / CTR-desynced burst
+            # decodes NEITHER (mask and status both None), so it still preserves state
+            # — the presence of a decoded frame is the distinguishing signal.
+            self.state.programs = {}
         return programs, mask
 
     async def get_programs(self) -> dict:
@@ -809,7 +818,12 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
                 )
                 await self.refresh_status()
                 _programs2, mask2 = await self._read_programs()
-                return bool((mask2 or 0) & bit) == on
+                # Route the confirm read through _enable_mask so a dropped #20 frame
+                # (mask2=None) fails safe in BOTH directions: `(mask2 or 0)` would let a
+                # DISABLE confirm True without verifying anything (0 & bit == 0 == "off"),
+                # asymmetric with enable. Reconstructing from last-known flags reports the
+                # unverified toggle as unconfirmed instead — the safe direction.
+                return bool(self._enable_mask(mask2) & bit) == on
             finally:
                 await self.connection.disconnect()
 
--- a/custom_components/orbit_bhyve/devices/protobuf.py
+++ b/custom_components/orbit_bhyve/devices/protobuf.py
@@ -180,13 +180,23 @@ def _build_set_epoch_time_pb(epoch_utc: int | None = None, tz_offset_sec: int | 
     #16 run-state, battery, clock, controller-mode, next-start), so it doubles as
     the coordinator-poll status elicitor while keeping the device clock honest
     (program start-times, rain-delay #3, and auto-close all key off this clock).
-    Defaults to the host's current time / UTC offset."""
-    now = datetime.now().astimezone()
-    if epoch_utc is None:
-        epoch_utc = int(now.timestamp())
-    if tz_offset_sec is None:
-        off = now.utcoffset()
-        tz_offset_sec = int(off.total_seconds()) if off is not None else 0
+    Defaults to HA's configured local time / UTC offset — NOT the container OS
+    timezone, which can diverge from the user's HA setting and would skew the
+    device's schedule next-start computation."""
+    if epoch_utc is None or tz_offset_sec is None:
+        try:
+            # HA's configured local time, not the container OS timezone. Local
+            # import so the module still loads without Home Assistant (the
+            # standalone protocol tests, which pass explicit epoch/offset).
+            from homeassistant.util import dt as dt_util
+            now = dt_util.now()
+        except ImportError:
+            now = datetime.now().astimezone()
+        if epoch_utc is None:
+            epoch_utc = int(now.timestamp())
+        if tz_offset_sec is None:
+            off = now.utcoffset()
+            tz_offset_sec = int(off.total_seconds()) if off is not None else 0
     body = _pb_field_varint(1, epoch_utc) + _pb_field_varint_signed(2, tz_offset_sec)
     return _pb_field_bytes(75, body)
 
@@ -680,6 +690,17 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
             finally:
                 await self.connection.disconnect()
 
+    def _enable_mask(self, mask: int | None) -> int:
+        """The current #20 enable bitmask to base a write on. A program read can
+        decode the #19 bodies but drop the trailing #20 mask frame (mask=None) on
+        a marginal link; falling back to 0 there would clear every OTHER slot's
+        enable bit on the next write. _read_programs already carries each slot's
+        enabled flag forward into state.programs, so reconstruct the mask from it
+        instead of collapsing to 0."""
+        if mask is not None:
+            return mask
+        return sum(1 << (sid - 1) for sid, s in self.state.programs.items() if s.enabled)
+
     @_ble_write_guard
     async def set_program(self, spec: ProgramSpec) -> bool:
         """Store a program (#19) and, if spec.enabled, drive the 3-write run
@@ -692,7 +713,7 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
             try:
                 letter = SLOT_LETTERS.get(spec.slot, str(spec.slot))
                 _programs, mask = await self._read_programs()
-                mask = mask or 0
+                mask = self._enable_mask(mask)
                 pb = _build_program_pb(spec)
                 await self.connection.send(_build_message(pb), drain_ms=2000)
                 self._stamp_command(f"program set {letter}", 0)
@@ -722,9 +743,14 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
                 await self.connection.send(
                     _build_message(_build_set_active_programs_pb(newmask))
                 )
-                # Confirm via #16.#9/#10 next-start (folded into state by the observer).
+                # Confirm via #16.#9/#10 next-start (folded into state by the
+                # observer). next_start_flags is a bitmask of whichever slot(s)
+                # start NEXT, so anchor the confirm to THIS slot's bit — otherwise
+                # an already-enabled program with an earlier start would confirm a
+                # slot that didn't actually arm. (A slot that armed but isn't the
+                # soonest reads as unconfirmed; the poll then re-reads real state.)
                 await self.refresh_status()
-                ok = bool(self.state.next_start_flags)
+                ok = bool(self.state.next_start_flags and self.state.next_start_flags & bit)
                 _LOGGER.log(
                     logging.DEBUG if ok else logging.WARNING,
                     "%s: %s program set %s enabled %s",
@@ -745,7 +771,7 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
                 _programs, mask = await self._read_programs()
                 bit = 1 << (slot - 1)
                 await self.connection.send(
-                    _build_message(_build_set_active_programs_pb((mask or 0) & ~bit))
+                    _build_message(_build_set_active_programs_pb(self._enable_mask(mask) & ~bit))
                 )
                 await self.connection.send(_build_message(_build_program_delete_pb(slot)))
                 await self.connection.send(_build_message(_build_set_timer_mode_pb(1)))
@@ -767,7 +793,7 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
                 return False
             try:
                 _programs, mask = await self._read_programs()
-                mask = mask or 0
+                mask = self._enable_mask(mask)
                 bit = 1 << (slot - 1)
                 newmask = (mask | bit) if on else (mask & ~bit)
                 await self.connection.send(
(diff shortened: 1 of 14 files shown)


exactmike:
Will address the failing checks with an update.

custom_components/orbit_bhyve/services.yaml
```diff
@@ -27,6 +27,128 @@ refresh_devices:
   name: Refresh devices from cloud
   description: Re-query the Orbit cloud for new/changed devices and updated network keys. Manual; no background polling.
 
+set_program:
+  name: Set watering program
+  description: Create or replace a watering program (slots A–D) on a B-Hyve valve (HT34A / HT25G2). Choose exactly one day-mode.
+  target:
+    device:
```
ljmerza: **B1 — Blocker (Hassfest):** device-target filters aren't accepted here.

```
[ERROR][SERVICES] Services do not support device filters on target, use a
device selector instead ... Got {'device': {'integration': 'orbit_bhyve'}}
```

The bare `integration:` under `device:` needs to move under the device selector's `filter:` key:

```yaml
  target:
    device:
      filter:
        - integration: orbit_bhyve
```

Same fix on `delete_program` (line 124) and `get_program` (line 140). The pre-existing `target:` in this file passes because it uses an `entity:` filter, which Hassfest still accepts as-is. (Worth confirming the exact selector shape against current HA `services.yaml` docs.)

custom_components/orbit_bhyve/strings.json
```diff
@@ -67,6 +67,36 @@
...
+      "name": "Set watering program",
+      "description": "Create or replace a watering program (slots A–D) on a B-Hyve valve (HT34A / HT25G2).",
+      "fields": {
+        "slot": {"name": "Program slot", "description": "Which program slot to write (A–D)."},
+        "day_mode": {"name": "Day mode", "description": "How often the program runs. Exactly one mode."},
+        "weekdays": {"name": "Weekdays", "description": "For day mode \"weekdays\" — which days to run."},
+        "interval_days": {"name": "Interval (days)", "description": "For day mode \"interval\" — run every N days."},
+        "interval_anchor": {"name": "Interval anchor date", "description": "For day mode \"interval\" — ISO date the interval counts from (default today)."},
+        "start_times": {"name": "Start times", "description": "One or more start times (HH:MM, device-local)."},
+        "zones": {"name": "Zones", "description": "Per-zone run time — a list of {zone, minutes}."},
```
ljmerza: **B2 — Blocker (Hassfest):** `{zone, minutes}` is parsed as a translation placeholder.

```
[ERROR][TRANSLATIONS] placeholders must be valid identifiers
([a-zA-Z_][a-zA-Z0-9_]*) ... Got 'Per-zone run time — a list of {zone, minutes}.'
```

Reword to drop the curly braces, e.g. `"Per-zone run time — a list of zone/minutes pairs."` The identical line in `translations/en.json:81` needs the same change (the two files are kept byte-identical).

custom_components/orbit_bhyve/devices/protobuf.py
```diff
@@ -469,3 +624,209 @@ async def _clear_rain_delay_unlocked(self) -> bool:
...
+        handshake (store -> enable #20 -> autoMode #14{1} -> re-send store+enable).
+        Returns True once the store read-back lands (store-only) or the device
+        reports a next-start (enabled). Never leaves the controller in offMode."""
+        async with self._api_lock:
+            if self.connection is None:
+                return False
+            try:
+                letter = SLOT_LETTERS.get(spec.slot, str(spec.slot))
+                _programs, mask = await self._read_programs()
+                mask = mask or 0
```
ljmerza: **C1 — Correctness (medium):** if the `#20` mask frame drops on this pre-write read, `mask or 0` collapses to `0`, and the `_build_set_active_programs_pb(newmask)` write below then clears every *other* slot's enable bit — silently disabling other programs on the device.

`_read_programs()` already carries each slot's `enabled` forward into `self.state.programs` for exactly this partial-read case, but the returned `mask` stays `None`. Consider reconstructing it from state when `None`:

```python
if mask is None:
    mask = sum(1 << (sid - 1) for sid, s in self.state.programs.items() if s.enabled)
```

Same `mask = mask or 0` pattern in `set_program_enabled` (line ~770), and the inline `(mask or 0)` in `delete_program`.

custom_components/orbit_bhyve/devices/protobuf.py
```diff
@@ -469,3 +624,209 @@ async def _clear_rain_delay_unlocked(self) -> bool:
...
+                    _build_message(_build_set_active_programs_pb(newmask))
+                )
+                await self.connection.send(_build_message(_build_set_timer_mode_pb(1)))
+                await self.connection.send(_build_message(pb), drain_ms=2000)
+                await self.connection.send(
+                    _build_message(_build_set_active_programs_pb(newmask))
+                )
+                # Confirm via #16.#9/#10 next-start (folded into state by the observer).
+                await self.refresh_status()
+                ok = bool(self.state.next_start_flags)
```
ljmerza: **C2 — Correctness (low-medium):** `#16.#9 next_start_flags` is a bitmask of whichever slot(s) start *next*, not necessarily this one. If another enabled program already has an earlier next-start, this returns `True` even when *this* slot didn't arm. Anchor the confirm to the slot bit:

```python
ok = bool(self.state.next_start_flags and self.state.next_start_flags & (1 << (spec.slot - 1)))
```

The current tests only exercise the single-program case, so this passes today.

custom_components/orbit_bhyve/devices/protobuf.py
```diff
@@ -117,6 +154,103 @@ def _build_rain_delay_pb(minutes: int, expiry: int | None) -> bytes:
...
+    """Clock-sync: #75 setEpochTime { #1 timeSecEpochUTC, #2 timezoneOffsetSec }.
+
+    This — NOT #18 setDateTime (a no-op over BLE, HW-verified) — sets the device
+    clock, and the app also uses it to trigger the first status burst on connect.
+    Its reply is content-identical to a #15 status (HW-verified 2026-07-06: same
+    #16 run-state, battery, clock, controller-mode, next-start), so it doubles as
+    the coordinator-poll status elicitor while keeping the device clock honest
+    (program start-times, rain-delay #3, and auto-close all key off this clock).
+    Defaults to the host's current time / UTC offset."""
+    now = datetime.now().astimezone()
```
ljmerza: **C3 — Correctness (low):** `datetime.now().astimezone()` reads the **container OS** timezone. The epoch (#1) is UTC-correct via `.timestamp()`, but the tz-offset (#2) can diverge from the user's configured HA timezone on container installs — and the device computes schedule next-starts from that offset. Prefer HA's own clock via `homeassistant.util.dt` (`dt_util.now()` for the aware local time, `.utcoffset()` for the offset).

ljmerza (commented):
Thanks for a thorough, well-documented PR. Overall this is solid, low-blast-radius work: strictly additive, gated behind `isinstance(device, BHyveProtobufDevice)`, mesh/hub paths untouched, with good self-contained test coverage and hardware-verified byte references.

Two framing notes:

- **CI is red** — Hassfest fails with two concrete errors (inline as **B1** / **B2** below). Blocks merge.
- **I can't hardware-verify the protobuf-family claims** — my own fleet is mesh HT25 only, so the fw0111/fw0107 byte references, the 3-write enable handshake, the `#75`==`#15` reply equivalence, and the latching identify all rest on your bench. Not asking you to re-prove them — just flagging that the byte-reference tests are what I'm leaning on.

**Blockers (must fix):** B1 (`services.yaml` device target) and B2 (`{zone, minutes}` placeholder), inline.

**Correctness worth addressing:** C1 (enable-bitmask clobber on a dropped `#20` frame — the one with real user impact), C2 (slot-specific enable confirm), C3 (OS vs HA timezone), inline.

**Minor / nits (non-blocking):**
- `get_program` calls `async_request_refresh()` right after the ephemeral `get_programs()` read → a second full BLE connect per device for a read-only service. Consider dropping it on the read path.
- `interval` day-mode doesn't validate that `interval_days` is present; `_build_program_pb` silently falls back to `interval_days or 1` (every-1-day), whereas `weekdays` mode validates. Worth a symmetric check in `_spec_from_call`.
- Reuse: `sensor.py` hardcodes `{"A":1,...}[letter]` and both `sensor.py`/`switch.py` re-declare `_UI_SLOTS`, while `PROGRAM_SLOTS`/`SLOT_LETTERS` already live in `base.py`.
- `run_program()` looks unused (no button/service calls it) — dead code or a future hook?
- `identify()` holds `_api_lock` for the full `seconds` (~6s), blocking polls/actuation on that device meanwhile — fine for a manual button, just noting.

**Not a defect, just a heads-up:** the PR description lists "Battery sensor no longer seeds from the cloud" as a behavioral change here, but that (`_RestoreLastValueSensor` in `sensor.py`) is already on `main` from #24 — this diff doesn't touch battery seeding. The genuinely new behavioral change is the `#75 setEpochTime` poll elicitor replacing `#15{}`, which is called out clearly.

exactmike:
Thanks — really useful review. Pushed fixes for everything; details below.

**Blockers (CI):**
- **B1** — moved the program services to a device-selector `filter:` list (`device: { filter: [ { integration: orbit_bhyve } ] }`) for `set_program` / `delete_program` / `get_program`. The entity-target service was already fine.
- **B2** — reworded the `zones` description to drop the `{…}` ("a list of zone/minutes pairs"), in both `strings.json` and `translations/en.json`.

Both were local Hassfest failures; the suite passes here and I expect CI green now.

**Correctness:**
- **C1** — good catch, this was the one with real user impact. Added `_enable_mask()`: when a read decodes the `#19` bodies but drops the `#20` frame (`mask=None`), it now reconstructs the mask from the last-known per-slot `enabled` flags (which `_read_programs` already carries into `state.programs`) instead of collapsing to `0`. Applied in `set_program`, `set_program_enabled`, and `delete_program`. Test: enabling slot D on a dropped-mask read no longer clears slot A's bit.
- **C2** — anchored `set_program`'s enable confirm to `next_start_flags & (1 << (slot-1))`, so an already-enabled program with an earlier start can't falsely confirm a slot that didn't arm. (A slot that armed but isn't the soonest now reads unconfirmed and the poll re-reads real state — the safe direction.) Added a multi-program test.
- **C3** — switched the `#75` tz-offset to HA's clock via `homeassistant.util.dt.now()` (local import, with a system-local fallback only for the standalone protocol tests, which have no HA). Epoch was already UTC-correct; this fixes the offset on container installs where the OS tz differs from the HA setting.

**Nits:**
- `get_program` no longer does a second BLE connect — `get_programs()` already refreshes `state.programs`, so it now pushes that with `async_set_updated_data()` instead of `async_request_refresh()`.
- Added the symmetric `interval_days` presence check in `_spec_from_call`.
- Removed `run_program()` — dead code (W0 found no immediate-run path, so it only aliased `set_program_enabled`).
- Hoisted `UI_SLOTS` into `base.py` next to `PROGRAM_SLOTS`/`SLOT_LETTERS`, dropped the per-platform `_UI_SLOTS` copies, and replaced the hardcoded `{"A":1,…}` in `sensor.py` with `PROGRAM_SLOTS`.
- `identify()` holding `_api_lock` for the flash duration: left as-is per your note (manual button; it's the intended latch-then-stop window).

**Battery heads-up** — you're right, thanks. That change (`_RestoreLastValueSensor`, no cloud seed) is already on `main` from #24; this diff doesn't touch battery. I've corrected the PR description to drop it — the only behavioral change here is the `#75` poll elicitor.

**On hardware:** understood, and no worries — the fw0111/fw0107 byte references, the 3-write enable handshake, the `#75`≡`#15` reply equivalence, and the latching identify are all bench-verified on my Gen2 + XD fleet, and the byte-reference tests pin them. Program get/set/delete + the switch toggles were verified end-to-end via HA on the fleet earlier; this review round's changes are covered by the unit tests, and I'll re-deploy and re-run the live CRUD pass over the fleet before I'd call it merge-ready.

exactmike:
Quick correction on **B1** and a CI update: my first pass (a `filter:` under `target.device`) was still wrong — current Hassfest rejects *any* device key under a service `target` (`raise_on_target_device_filter`), whatever the shape. So I dropped the `target:` block on the three program services and exposed the device as a required **`device_id` field with a device selector** (`filter: [integration: orbit_bhyve]`) — which is the data key the handler already reads. Mirrored the field in `strings.json`/`en.json`.

**CI is green now** (Hassfest + HACS both pass). Everything else from the review round is in the three commits above.

ljmerza:
Reviewed the full diff against `main`. Impressive work overall — the `_enable_mask` reconstruction fallback, the documented no-retry rationale, and the CTR-encrypted multi-frame reassembly test are all excellent. I verified the usual silent-breakage spots (imports, `_on_notify` dedup ordering, `_pb_field_varint_signed` precedence, `split_inner_messages` length math, `_drain` timing semantics) and they all check out. A few findings, most important first:

### 1. Probable bug — "Next run" never clears once the device stops reporting it (`devices/status.py`)

`apply_status_plaintext` only updates the next-start state when the field is present:

```python
if st.next_start_flags is not None:
    device.state.next_start_flags = st.next_start_flags or None
    ...
```

Protobuf encoders normally omit zero-valued scalar fields, so when the last enabled program is disabled/deleted, `#16.#9` is likely **absent** from the status — leaving the Next-run sensor frozen at the old timestamp (and `next_start_flags` stale) until an HA restart. The rain-delay handling a few lines up has an explicit clear-on-absence `elif` branch for exactly this failure mode (the "7 hours ago" bug comment).

It also enables a narrow false confirm in `set_program`: `ok = bool(self.state.next_start_flags and self.state.next_start_flags & bit)` reads state, so stale flags with this slot's bit already set would pass even if the fresh `#16` omitted `#9`.

Worth a hardware check: does the device emit `#9 = 0` explicitly when nothing is armed? If not, this needs a clearing branch symmetric to the rain-delay one.

### 2. Real footgun — YAML parses unquoted start times as sexagesimal ints (`_spec_from_call` in `__init__.py`)

Verified locally with PyYAML:

```python
>>> yaml.safe_load('[06:00, 18:00]')
['06:00', 1080]   # 18:00 → int 1080; 06:00 survives only because of the leading zero
```

`cv.string` coerces `1080` → `"1080"`, then the parser computes `1080*60 % 1440 = 0` — so `start_times: [06:00, 18:00]` silently schedules 06:00 and **midnight**. Also, `HH:MM:SS` strings (what HA time selectors produce) raise a raw `ValueError` on `int("00:00")`.

Suggest: validate tokens with a proper time parse (or `cv.time`), accept `HH:MM[:SS]`, and reject values ≥ 1440 instead of wrapping with `% 1440`.

### 3. Minor — disable-path false confirm in `set_program_enabled` (`devices/protobuf.py`)

```python
_programs2, mask2 = await self._read_programs()
return bool((mask2 or 0) & bit) == on
```

If the confirm read drops the `#20` mask frame (`mask2=None`), a **disable** returns `True` without verifying anything, while an enable conservatively fails — asymmetric. Reusing `self._enable_mask(mask2)` for the check would fix it.

### 4. Minor — `get_programs()` is unguarded on the service path

Writes get `_ble_write_guard` and the poll path try/excepts, but the `get_program` service calls `get_programs()` raw — a connect failure surfaces a raw `BleakError` to the service caller. Wrapping in `HomeAssistantError` would be cleaner UX (or accept as intentional visibility for an on-demand read).

### 5. Edge — `_read_programs` never clears state on a body-less dump

The `if programs:` gate means that if the hardware omits empty slots from the `#10` dump (rather than echoing NotSet bodies) and **all** programs get deleted via the app, `state.programs` retains the stale schedule forever. Moot if the dump always includes per-slot NotSet bodies — worth confirming on hardware alongside item 1.

### Nits

- `get_program`'s response is keyed by `coord.device.name` — two same-named devices overwrite each other; keying by cloud_id/mac with a `name` field would be robust.
- `identify()`: if the stop write raises, `_ble_write_guard` swallows it and the latched LED keeps flashing. It does warn, so acceptable as best-effort.
- Heads-up: this PR and #32 both touch `PLATFORMS` in `__init__.py` and append to `tests/test_devices.py` — whichever merges second needs a trivial rebase. They compose correctly otherwise (the `_is_status_elicitor` helper here already treats both `#15` and `#75`, and #32's confirm reads intentionally keep the bare `#15`).

exactmike:
Thanks — another sharp review. Pushed fixes for all of it (`6a1e9f4`), and I
re-deployed the branch to the fleet and ran a live CRUD pass on BTValve01
(HT25G2 fw0111). CI is green (Hassfest + HACS). Details, most important first —
including a couple of hardware answers to your open questions.

### 1. "Next run" clear-on-absence — fixed, + hardware answer

Added the clear branch symmetric to the rain-delay one: when a `#16` status
decodes but carries no `#9`, `next_start_flags`/`next_start_at` are cleared (so
the sensor can't freeze and `set_program` can't confirm off stale flags).

Hardware answer to your "does the device emit `#9 = 0` explicitly?": **yes, it
does** — on fw0111 the `#16` status always carries `#9` (`4800` when nothing is
armed, `4801`/`4809` when a slot is). So the pre-existing first branch
(`next_start_flags or None`) was already clearing it, and the frozen-sensor bug
doesn't actually reproduce on this hardware. The new branch is therefore a
**defensive guard for the omit case** you flagged (which is the normal protobuf
behavior and may well be what other firmware does) rather than a fix for an
observed break here. Verified live regardless: disabling the last enabled
program clears the Next-run sensor; re-enabling repopulates it. Unit test covers
the omit path (a `#16` with no `#9` after an armed state).

### 2. YAML sexagesimal start-times — fixed (this was the real one)

Replaced the `(int(h)*60+int(m)) % 1440` parse with a real `HH:MM[:SS]` parser
(`parse_start_minutes`, kept HA-free so it unit-tests standalone) that rejects
out-of-range / non-time tokens instead of wrapping, and maps the failure to a
`ServiceValidationError`. Verified live on BTValve01: `start_times: ["06:00",
"18:00"]` encodes to the wire packed-varint `[360, 1080]` and reads back
`["06:00","18:00"]` (18:00, not midnight); `start_times: ["1080"]` now raises
`ServiceValidationError: start_times entries must be HH:MM (or HH:MM:SS); got
'1080'. Quote bare times in YAML…` with no BLE traffic. Unit tests for both.

### 3. Disable-path false confirm — fixed

`set_program_enabled` now runs the confirm read through `_enable_mask(mask2)`
instead of `(mask2 or 0)`, so a dropped `#20` frame reads as *unconfirmed* in
both directions (a disable no longer returns `True` without verifying). Unit
test: disable with `mask2=None` no longer confirms.

### 4. `get_program` unguarded read — fixed

Wrapped the `get_programs()` call so a `BleakError`/timeout becomes a
`HomeAssistantError` with context. This one actually fired live during the CRUD
pass (a transient connection-drop mid-read over the ESPHome proxy) and surfaced
as `could not read programs from BTValve01: Peripheral … changed connection
status …` instead of a raw bleak traceback — the intended UX.

### 5. `_read_programs` never clears on a body-less dump — fixed, + hardware answer

Added: when the `#10` dump decodes a coherent frame (a `#20` mask and/or `#16`
status) but zero `#19` bodies, `state.programs` is cleared; a truncated /
CTR-desynced burst decodes neither and still preserves last-known.

Hardware answer to your "worth confirming on hardware": on fw0111 the `#10` dump
**echoes a per-slot NotSet body for all six slots** (`9a0104 08NN 1200`) rather
than omitting deleted ones — so the existing `if programs:` overwrite already
clears stale state, and, like item 1, the stale-retention bug doesn't reproduce
here. Verified live: deleting the last program flips every Program A–D sensor to
`empty` and the switches to `unavailable`. The new branch is again a defensive
guard for firmware that *does* omit. Both findings 1 & 5 are provably inert on
this hardware (the branches only fire on the omit case) — no regression risk —
and are locked by unit tests.

### Nits

- **n1** — `get_program` response is now keyed by device MAC with a `name`
  field, so two same-named devices can't clobber each other.
- **n2** — left `identify()`'s stop as best-effort (it warns), per your note.
- **n3** — understood on the `PLATFORMS` / `tests/test_devices.py` overlap with
  #32; I'll rebase whichever of the two merges second.

Test count is up with the new regressions and they're green here; I proved each
new behavioral test fails on the pre-fix code before committing.

PR #32, opened 2026-07-11
fix(rain-delay): don't self-overwrite on units that ignore #15; add a rain-delay preset select

exactmike:
## Summary

Two rain-delay changes for the protobuf family (HT34A XD + HT25G2 Gen2):

1. **Bug fix — `set_rain_delay` / `clear_rain_delay` could silently cancel the change
   they just made.** On a unit that ignores the passive `#15` `getDeviceStatus` query,
   the confirm read-back falls into `_recover_status`, whose status elicitor is a
   `setRainDelay` (`#17`) re-assert of `self.state` (`_noop_rain_delay_pb`). Because the
   device suppresses the set reply while it is active, `self.state.rain_delay_minutes`
   still held the **pre-write** value at that point, so the no-op sent `#17{#1=0}` and
   **cleared the delay just set** (and, symmetrically, re-asserted a stale delay to
   **undo a clear**). The device then echoed `0`, the entity snapped back to `0`, and the
   log showed `rain-delay set … unconfirmed`. Hardware-observed on a live valve.

2. **Feature — a rain-delay preset `select` entity.** Exposes the delay as coarse
   hour/day presets so one pick is exactly **one BLE write**, with no ambiguity about
   when the value is sent (unlike a number stepper/slider, which writes on every
   click/drag and can queue intermediate writes that contend with the coordinator poll
   for the device's single BLE session).

### `0232197` — fix + regression test
- In `set_rain_delay` and `_clear_rain_delay_unlocked`, **seed the intended rain-delay
  state onto `self.state` before the confirming read-back**, so a `_recover_status`
  no-op re-asserts the *new* value (idempotent with the write) and the device echoes the
  real `#16.#13` back — instead of reverting to stale pre-write state.
- **Gate "confirmed" on `_status_parsed`** so the seed alone can't report a false confirm
  when no `#16` was decoded.
- New regression test (`tests/test_devices.py`) models a unit that ignores `#15` and
  whose set reply is suppressed; it **reproduces the exact `unconfirmed` symptom without
  the fix** and passes with it.

### `7b92cc1` — preset select
- New `select.py` → one `BHyveRainDelaySelect` per protobuf-family device. Options:
  `Off, 1/2/4/8/12/16 hours, 1/2/3/5/7 days`. `current_option` reflects the live
  `#16.#13` state (a non-preset active value shows nothing selected rather than
  mislabeling); selecting maps to `set_rain_delay` / `clear_rain_delay`, one write.
- `__init__.py`: add `Platform.SELECT` to `PLATFORMS`.

## Backwards compatibility

Strictly additive.
- The fix only changes the ordering/gating inside the existing rain-delay methods; the
  wire protocol and the happy path (a unit that answers `#15`) are unchanged.
- The existing **"Rain delay" number** entity (arbitrary hours) is untouched and stays
  alongside the new dropdown, so existing dashboards / automations keep working.
- The `select` platform is gated to `BHyveProtobufDevice`; mesh/hub devices are unaffected.

## Testing

- **Unit:** 103 tests green. The new regression test fails without the fix
  (`rain-delay set 120m unconfirmed`, minutes wiped to 0) and passes with it.
- **CI:** Hassfest + HACS both pass on the fork's `validate.yaml`.
- **Hardware (fleet, deployed to HA):** set/cleared rain delay across BTValve01–04 (Gen2)
  and the BT4ValveXD01 (HT34A) from the new dropdown — each pick is a single
  `rain-delay set …m confirmed`, no `unconfirmed` warnings, and the dropdown reflects the
  device's live state (e.g. a 1440 m delay reads back as "1 day"). The pre-fix symptom
  (entity snapping back to 0) is gone.

---

### DIFF
--- a/custom_components/orbit_bhyve/__init__.py
+++ b/custom_components/orbit_bhyve/__init__.py
@@ -46,6 +46,7 @@ PLATFORMS: list[Platform] = [
     Platform.BINARY_SENSOR,
     Platform.NUMBER,
     Platform.BUTTON,
+    Platform.SELECT,
 ]
 
 
--- a/custom_components/orbit_bhyve/devices/protobuf.py
+++ b/custom_components/orbit_bhyve/devices/protobuf.py
@@ -438,10 +438,21 @@ class BHyveProtobufDevice(BHyveBleDeviceBase):
                 plaintext = _build_message(_build_rain_delay_pb(minutes, expiry))
                 notifs = await self.connection.send(plaintext, drain_ms=2000)
                 self._stamp_command(f"rain_delay set {minutes}m", len(notifs))
+                # Seed the INTENDED delay onto state before the confirm read. If that
+                # read falls into _recover_status (a unit that ignores the passive
+                # #15), its status elicitor is a setRainDelay re-assert of self.state
+                # (_noop_rain_delay_pb). Left at the pre-set value it would re-send
+                # #17{#1=0} and CLEAR the delay we just set (the reported bug). Seeding
+                # the new value makes that re-assert idempotent with the set, so the
+                # device echoes the real #16.#13 back and confirmation stays honest.
+                self.state.rain_delay_minutes = minutes
+                self.state.rain_delay_ends = datetime.fromtimestamp(expiry, tz=timezone.utc)
                 # Read back the authoritative #16.#13 echo via #15{} rather than trusting
-                # the set reply's push (which the device suppresses while active).
+                # the set reply's push (which the device suppresses while active). Gate
+                # "confirmed" on a real #16 decoding this cycle (_status_parsed) so the
+                # seeded value alone can't report a false confirm when no status came back.
                 await self.refresh_status()
-                ok = bool(self.state.rain_delay_minutes)
+                ok = self._status_parsed and bool(self.state.rain_delay_minutes)
                 _LOGGER.log(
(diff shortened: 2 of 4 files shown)


ljmerza:
Reviewed the full diff against `main` — **no bugs found**.

Verified the spots most likely to break silently:

- `datetime`/`timezone` are already imported in `devices/protobuf.py`, so the seed lines compile.
- `expiry` is always an int at the seed site (`expiry = base + minutes * 60`), so `datetime.fromtimestamp(expiry, tz=timezone.utc)` is safe.
- The public `clear_rain_delay()` the new select calls exists on `main`.
- `_status_parsed` is only set when a real `#16` run-state decodes (`apply_status_plaintext`), so the seeded value genuinely cannot self-confirm — the confirm gating is sound.
- The regression test reproduces the exact pre-fix failure (recovery no-op re-asserting stale state) and the preset→minutes table is correct.

Only note: on an *unconfirmed* set, `self.state` keeps the optimistic seeded value until the next poll corrects it — reasonable trade-off given the write did go out.

Heads-up: this PR and #31 both touch `PLATFORMS` in `__init__.py` (SELECT vs SWITCH) and append to `tests/test_devices.py`, so whichever merges second needs a trivial rebase. They compose correctly otherwise — the confirm reads here intentionally keep the bare `#15`, and #31's test helper already treats both `#15` and `#75` as status elicitors.

PR #36, opened 2026-07-17
feat(battery): per-device battery chemistry select so the % gauge matches non-alkaline AA cells

exactmike:
## Summary

The battery percent is a linear voltage gauge with **alkaline** endpoints (0% @ 2400 mV,
100% @ 3000 mV). That's accurate for the alkaline cells Orbit assumes, but it reads other AA
chemistries badly — most importantly **Ni-MH rechargeables**, which spend ~80% of their life
on a flat ~2.4–2.5 V plateau. On that plateau the current formula reports a low percent
(and trips Home Assistant's low-battery alerts) while most of the charge remains. Live example
from my fleet: a Ni-MH valve reading **2628 mV** shows just **38% remaining** on the alkaline
curve, when the correct Ni-MH curve puts it at about **70% remaining** — i.e. it looks nearly
flat while it's actually well over half full.

Chemistry can't be inferred from a single stateless voltage reading (2550 mV could be alkaline
at ~25% or Ni-MH at ~80%, and AA cells are charged off-device so there's no charging signal), so
this adds a **per-device "Battery chemistry" select** (config category) that picks the
voltage→percent curve. It defaults to **Alkaline**, so out-of-the-box behavior is unchanged.

## Changes (`f639295`)

- **`devices/base.py`**
  - `BATTERY_CHEMISTRIES`: chemistry → `(mV@0%, mV@100%)` endpoint map, plus
    `DEFAULT_BATTERY_CHEMISTRY = "alkaline"`.
  - `_mv_to_pct(mv, chemistry="alkaline")` gains an optional chemistry arg (defaults to
    alkaline, so every existing caller is unchanged); unknown chemistry falls back to alkaline.
  - `battery_pct` becomes a **computed property** of `battery_mv` + `battery_chemistry` instead
    of a stored attribute, so changing the chemistry re-gauges immediately with no new BLE read.
    (`battery_pct` was already just `_mv_to_pct(battery_mv)` at every write site, so this removes
    redundant state rather than adding any.)
  - New `battery_chemistry` instance attribute (default alkaline), set by the select.
- **`select.py`** — `BHyveBatteryChemistrySelect` on every BLE device (`connection is not None`;
  mesh + protobuf). `RestoreEntity`, `EntityCategory.CONFIG`. Persists the pick across restarts
  via restore-state (the device can't report its own chemistry), and on change recomputes the
  percent sensor via `async_update_listeners()` — no device round-trip.
- **`sensor.py` / `devices/status.py`** — consume the property; drop the now-redundant stored
  percent write and the dead `_mv_to_pct` fallback branch + import.
- **`README.md`** — document the selector and the regulated-Li-Ion caveat (see below).
- **`tests/test_devices.py` / `tests/test_protocol.py`** — chemistry curve endpoints
  (Ni-MH / lithium / regulated / unknown-fallback) and a `battery_pct`-follows-chemistry
  regression that fails on the pre-change code; `apply_status_plaintext` now stores mV only, so
  the protocol test asserts `battery_mv` and the percent is covered as a property.

## Proposed voltage endpoints (2× AA series)

| Chemistry | 0% | 100% | Rationale |
| :--- | :---: | :---: | :--- |
| Alkaline (default) | 2400 | 3000 | Orbit's stock curve; unchanged |
| Ni-MH rechargeable | 2350 | 2750 | Flat ~2.4–2.5 V plateau; conservative 0% floor (see safety note) |
| Lithium primary (Li-FeS₂) | 2600 | 3400 | High, very flat ~3.0 V plateau, sudden end-cliff |
| Lithium regulated 1.5 V | 2400 | 3000 | Constant output → unmeasurable by percent (see caveat) |

**Regulated 1.5 V Li-Ion caveat (documented in the README):** these cells hold a constant output
voltage until their internal cell hits cutoff and the regulator shuts off instantly, so *no*
percentage gauge can track them — the percent reads full right up until the valve dies. The
selector still lets users label them; the README tells them to watch the voltage sensor instead.

**Ni-MH safety note:** the 2350 mV 0% floor is deliberately above the cell's true-empty
(~2.1–2.2 V) so the low-battery warning fires *before* the closing solenoid pulse can sag too low
to latch the valve shut under water pressure.

## Backwards compatibility

Strictly additive.
- Default chemistry is Alkaline, and `_mv_to_pct`'s new arg defaults to alkaline, so the percent
  sensor behaves exactly as before for anyone who doesn't touch the new selector.
- The select is added on every BLE device but sits in the **config** entity category, so it stays
  out of the default dashboard cards.
- No wire-protocol changes; no new BLE traffic (chemistry is pure HA-side state).

## Testing

- **Unit:** 162 tests green. New tests cover each chemistry's curve and the
  chemistry-change-re-gauges behavior (the latter fails on the pre-change code).
- **CI:** Hassfest + HACS both pass on the fork's `validate.yaml`.
- **Hardware (fleet, deployed to HA):**
  - Select appears as a config entity on all 6 BLE devices; the pick survives a restart
    (restore-state).
  - Switching a Gen2 valve Alkaline→Ni-MH re-gauged live with no BLE round-trip: 2628 mV went
    from **38% → 70%**.
  - **Alkaline cross-checked against the B-Hyve app's own gauge:** XD at 2828 mV reads **71%** on
    the alkaline curve, matching the **71%** the B-Hyve app showed for the same device — so the
    default curve reproduces Orbit's gauge, not just my old formula.

## Note for review (honest provenance)

The **alkaline** curve is cross-checked against Orbit's own displayed gauge (above). The **Ni-MH**
and **lithium** endpoints are conservative estimates from published discharge profiles, **not**
calibrated against a known state of charge — there's no cheap ground truth for them (Orbit's cloud
gauge assumes alkaline and mis-reports them the same way). They're chosen to be directionally
right and fail-safe rather than exact, so I'm happy to adjust the values if you have better data.

On the shape: this is deliberately a **per-device select**, not an integration-wide option. A
single global setting would skew every device the moment one valve runs a different chemistry
(e.g. Ni-MH in the Gen2 valves, alkaline in an XD timer), which is the mixed-fleet case
this allows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,10 +47,22 @@ Per discovered sprinkler device:
   **optimistic** (derived from the last command, not from a decoded
   device status).
 - **Battery (%)** sensor — live, BLE-sourced. Decoded from the device's
-  info-ack frame on every poll, no cloud round-trip after setup.
+  info-ack frame on every poll, no cloud round-trip after setup. The percent
+  is a linear voltage gauge whose discharge curve is chosen by the **Battery
+  chemistry** selector below.
 - **Battery voltage (mV)** sensor — same source as the percent sensor;
   disabled by default, enable it from the entity's settings if you want
   the raw reading.
+- **Battery chemistry** (`select`, config category) — the AA cell chemistry
+  you installed (**Alkaline** / **Ni-MH rechargeable** / **Lithium primary**
+  / **Lithium regulated 1.5 V**), which selects the voltage→percent curve.
+  Chemistry can't be auto-detected from a single voltage reading (2550 mV
+  could be alkaline at 25% or Ni-MH at 80%), so it defaults to Alkaline
+  (Orbit's stock assumption) and you set it per device; changing it re-gauges
+  the percent immediately, no device round-trip. **Note:** regulated 1.5 V
+  Li-Ion cells output a constant voltage until they abruptly cut off, so no
+  percentage gauge can track them — the percent reads full right up until the
+  valve dies. Watch the voltage sensor instead.
 - **Signal strength (RSSI)** sensor — the BLE advertisement RSSI from
   Home Assistant's bluetooth manager (works even while disconnected);
   disabled by default.
--- a/custom_components/orbit_bhyve/devices/base.py
+++ b/custom_components/orbit_bhyve/devices/base.py
@@ -15,11 +15,41 @@ from ..const import DEFAULT_FLOW_COUNTS_PER_GALLON
 _LOGGER = logging.getLogger(__name__)
 
 
-def _mv_to_pct(mv: int) -> int:
-    """Linear approximation matching the cloud's discharge curve to within
-    a few percent: 0% at 2400 mV, 100% at 3000 mV. Tuned against three
-    live devices (Hill 33%/2602 mV, Corner 34%/2606 mV, Deck 65%/2771 mV)."""
-    pct = round((mv - 2400) * 100 / 600)
+# Per-chemistry linear fuel-gauge endpoints: (mV at 0%, mV at 100%). AA cells
+# in a 2x series (3 V nominal). See the "Battery Chemistry Options" plan for the
+# electrochemical rationale; the conservative Ni-MH 0% floor (2350 mV, well above
+# true-empty) exists to warn before solenoid-latch failure under water pressure.
+# Regulated 1.5 V Li-Ion outputs a constant ~3000 mV until it dies, so no linear
+# gauge can track it — it shares the alkaline endpoints and is documented as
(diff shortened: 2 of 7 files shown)


ljmerza (approved):
Two-axis review passed: Standards — 0 hard violations, correct RestoreEntity/CoordinatorEntity MRO, correct async_update_listeners() vs async_request_refresh() usage, no orphaned battery_pct writes. Spec — 10/10 requirements met, arithmetic and unknown-chemistry fallback verified. Minor non-blocking notes: restore path doesn't force an immediate re-gauge (harmless under normal setup ordering), and a pre-existing rain-delay docstring reflow. Approving.

PR #43, opened 2026-07-26
feat: add Gen1 flow measurement for HT25 hose-tap timers

FabsFabios:
Adds **Flow rate** and **Water used** sensors for HT25 (fw0085) hose-tap
timers, decoded from the timer's own inline flow meter over an undocumented BLE
subscription. Per-device opt-in, default off.

## Protocol

Subscribe: type `0x89` / seq `0x0E`, payload
`struct.pack("<HH", interval_ms, sample_count)`.

Samples: seq `0x0B`,
`[mesh:2][push_ctr:1][seq:1][routing:1][rate u16][elapsed u16][cumulative u16]`,
little-endian.

- `pt[2]` — push message counter cycling `0x80`–`0xBF`, shared across flow,
  status and battery pushes. **Not** a type byte, and the `0x40` reply bit is
  clear across that whole range, so neither a type check nor a TX-echo guard is
  possible on this frame.
- `pt[5:7]` — the device's own integer ticks/second. Ignored: at 118 counts/L
  one step is 0.51 L/min, ~3x coarser than deriving it from the counter.
- `pt[7:9]` — device-side seconds since START. Advances by exactly the
  subscribed interval and survives a mid-run re-subscribe. Used as `dt`, which
  keeps the host clock out of the decode entirely.
- `pt[9:11]` — cumulative ticks, u16, reset per run.

Run end: seq `0x0C`, run duration at `pt[6:8]` LE. Not part of the subscription
budget — observed arriving 46 s after a subscription had expired — so it stays
reliable as a release trigger. Its trailing field is `duration x integer` on
every capture and disagrees with the measured total by ~2.2%; unidentified and
unused.

Both subscribe fields are honoured, but the subscription also carries a ~300 s
ceiling: the device stops after `count` samples **or** ~300 s, whichever comes
first. Measured on four units:

| request | samples | span |
|---|---|---|
| 1000 ms x 10 | 10 | 10 s |
| 3000 ms x 700 | 97 | ~300 s |
| 1000 ms x 700 | 297 | ~300 s |

After a run ends the device keeps sampling with every field zeroed for as long
as the subscription lives.

## Design

**Session hold.** `_arm_idle_timer` fires only on writes, and a flow
subscription writes once then only receives — so at the default 60 s idle the
link closes mid-stream and a 10-minute program records roughly 28% of its
water. `hold_open()`/`release()` on `BHyveBleConnection`: idempotent,
self-expiring on `duration + grace`, re-arming the idle timer rather than
disconnecting (keeping one place in the code that closes sessions), and cleared
on disconnect so a stale hold can't suppress the next connection's timer.

**Renewal.** Re-subscribe every 240 s while watering, inside `_api_lock`,
skipping rather than queueing so a renewal can't fire late into a session STOP
has already closed. Verified on hardware: 412 samples, six renewals, gapless
over 20.6 minutes.

**Release** on six paths — seq `0x0C`, `async_close_valve`, the coordinator's
wall-clock auto-close (via a new `on_watering_finished()` hook: a run ending on
the device's own timer never reaches the valve entity), `stop_all`, config
entry unload, and the switch being turned off mid-run.

**Opt-in, default off.** Measuring holds the timer's single BLE session for the
whole program — more radio-on time than `DEFAULT_MESH_STATUS_POLL`, and it locks
the B-hyve app out for the duration. Entities are created whenever the model has
the meter, since nothing can create entities after `async_setup_entry` (the
options flow deliberately doesn't reload), and report `unavailable` while off —
mirroring `BHyveProgramEnableSwitch.available`. The calibration number stays
settable so a known value can be entered before the first measured run.

**Volume** reads the cumulative counter with a wrap carry rather than summing
deltas. The counter resets per run, so the first sample's ticks are real water;
the previous approach discarded them as a baseline. The u16 wraps at ~27 min on
a 40 tick/s zone, inside the supported program length.

## Calibration

`number.<device>_flow_calibration`, counts/litre, per device, `RestoreNumber`.
Four fw0085 units measured **113–118** against the vendor app; default 112.

Method: run the same duration twice on one tap, once from the B-hyve app and
once from Home Assistant, then `new = current x (HA litres / app litres)`. Back
to back, so mains pressure doesn't drift between them.

## Testing

219 tests (186 upstream + 33 new across two files). No Home Assistant required.
Async work uses `asyncio.run()` inside sync tests rather than
`@pytest.mark.asyncio`, since pytest-asyncio isn't in `requirements-test.txt`
and a marker-based test would break under a clean install.

Both critical tests are mutation-verified — the fix reverted, the test confirmed
to fail, the fix restored:

- `test_async_unload_cancels_the_renewal_task` — the orphaned-task regression
- `test_hold_open_disarms_the_idle_timer_and_blocks_re_arming` — the hold guard

The decode tests replay real capture values: six consecutive frames from a
hardware log land on 104.80 L and 17.5 L/min, matching the vendor app's figure
for that timer.

## Limitations

- **Only runs started through Home Assistant are measured**, including the
  `start_watering` service. A run from the timer's physical button or the
  B-hyve app produces no flow data.
- Only fw0085 tested.


ljmerza (changes requested):
Thanks for this — the reverse-engineering write-up is excellent, and the decode itself (wrap-safe u16 accumulation, the increment and `dt` sanity windows, mesh-address filtering against neighbour timers) is the right shape and properly tested. The `mesh_device_id is None` early return before touching `mesh_address` is exactly the right call given `_observe_plaintext` runs on every inbound frame. And recomputing volume from raw `_gen1_total_ticks` on each sample — so recalibrating mid-run retroactively corrects the whole run's total — is a genuinely nice touch.

I applied the patch to a clean checkout and ran the suite. A few things need addressing before merge.

## Blocking

### 1. Two of the new tests error out

`main` gives `181 passed`. This branch gives `197 passed, 2 errors`:

```
ERROR tests/test_gen1_flow.py::test_rate_computed_inside_dt_window
ERROR tests/test_gen1_flow.py::test_rate_suppressed_outside_dt_window_but_volume_still_books
RuntimeError: generator raised StopIteration   (in verify_cleanup teardown)
```

Both reproduce running a single test on its own, so it isn't cross-test pollution.

Cause: the tests patch the **global** clock with a two-element iterator —
`monkeypatch.setattr(time, "monotonic", lambda: next(ticks))`. `pytest_homeassistant_custom_component`'s
autouse `verify_cleanup` fixture calls `time.monotonic()` during teardown, after the iterator is
exhausted, and `StopIteration` inside a generator becomes `RuntimeError`.

Quick fix is a non-exhausting fake (`lambda: next(ticks, 102.0)`). Better: read the clock from a
module-level `_now()` in `ht25.py` so tests patch this integration's clock rather than the stdlib's.

(The description mentions 202 passing tests — I get 199 as the expected total, 181 + 18 new.)

### 2. Duplicate `set_plaintext_observer` in `connection.py`

The diff adds a second definition of a method that already exists at `connection.py:119`. Python
silently keeps the last one, so the existing `PlaintextObserver | None` annotation and the docstring
explaining the rx-counter desync rationale are dropped. The bodies are identical, so nothing breaks —
it looks like an accidental paste. `connection.py` shouldn't need any change for this feature; suggest
reverting the file entirely.

### 3. `_observe_plaintext` loses its docstring

`self.apply_gen1_flow_frame(pt)` was inserted *above* the docstring, which demotes it to a no-op string
expression — `BHyveHT25Device._observe_plaintext.__doc__` is now `None`. That docstring is the only
written record of the ack-frame layout and the off-timer rationale. Moving the call below it (and below
`super()._observe_plaintext(pt)`, matching the existing ordering) fixes it.

### 4. Flow rate never returns to zero

`state.flow_lpm_gen1` is only ever assigned inside `if 0 < inc < 30000`, so once flow stops the last
non-zero rate sticks until the *next* run calls `reset_gen1_flow()`. Feeding the patched code a run
followed by five static samples:

```
while flowing      -> rate=20.0  vol=3.0
after flow STOPS   -> rate=20.0  vol=3.0
```

`stop_flow_subscription` doesn't clear it either, so the sensor shows a phantom 20 L/min between runs.
Setting `flow_lpm_gen1 = 0.0` on `inc == 0` (with `dt` in the sanity window), and clearing it on
unsubscribe, should cover it.

### 5. Frame match is looser than the rest of the file

`apply_gen1_flow_frame` accepts anything where `pt[0:2]` matches the mesh address and `pt[3] == 0x0b`.
It doesn't check the type byte (`0x8B`) or the routing byte (`pt[4] == D747_ROUTING`), and has no
TX-echo guard — whereas `_observe_plaintext` deliberately checks both plus the `0x40` reply bit. Any
frame carrying seq `0x0B` will decode as a flow sample. Suggest tightening the guard and adding
`SEQ_FLOW = 0x0B` to the `SEQ_*` block rather than inlining the literal.

### 6. Flow subscribe bypasses `_api_lock`

`start_watering` holds `self._api_lock` across its whole connect → command → disconnect session, and
the comment there is explicit: *"an active status poll (`_connect_refresh`) must never interleave with
an actuation on the device's single BLE session."*

`start_flow_subscription` spawns a bare `hass.async_create_task` that calls `connection.send(...)`,
reconnecting **outside** that lock — concurrently with the `await self.coordinator.async_request_refresh()`
fired one line earlier in `async_open_valve`. `connection._lock` serialises the individual write but not
the session. Sending the subscribe frame from inside `start_watering`'s existing locked block, right
after START, would sidestep this entirely and drop the background task.

## Non-blocking follow-ups

- **`_loop()` isn't a loop** — it sends one frame and returns, so by the time `async_close_valve` runs
  the task is finished and `.cancel()` is always a no-op. No unsubscribe frame is sent, and the task
  isn't tracked at config-entry unload.
- **`has_flow_gen1` reinvents the `has_flow` pattern.** The repo already does this: `has_flow: bool = False`
  on `BHyveBleDeviceBase`, `has_flow = True` on the subclass, plain attribute access in `diagnostics.py`.
  This PR declares it only on HT25 and reads it via `getattr(..., False)` in four places, with both new
  class attributes at the bottom of the class body rather than up with `frame_magic`/`trailer_const`.
  Worth adding to the diagnostics payload too.
- **Gen1 "Water used" doesn't restore** across HA restarts, unlike the Gen2 `BHyveWaterUsedSensor` which
  extends `_RestoreLastValueSensor`. Combined with `TOTAL_INCREASING`, resetting to `None` per run reads
  to long-term statistics as a gap rather than a meter reset — `0.0` would be truer.
- **`number.py` setup doesn't skip `connection is None`** the way `sensor.py` does, so a key-less HT25
  record gets a calibration entity with no flow sensors to calibrate.
- **Function-local imports** — `import time` and `import asyncio` inside method bodies while `struct` went
  to the module top. `asyncio` is already imported at the top of `ht25.py`.
- **Clamp literals** — `max(50, min(500, ...))` twice, duplicating `_attr_native_min_value` /
  `_attr_native_max_value`. `MIN_MINUTES` / `MAX_MINUTES` are the existing precedent.
- **Unrelated churn** worth reverting to keep the diff focused: `_LOGGER` moved between `import logging`
  and `from typing import Any` in `valve.py` (splitting the import block), removed blank lines in
  `valve.py` / `sensor.py` / `number.py`, and the leftover development logging in `async_open_valve`
  (the `"MISSING"` sentinel and `"calling start_flow_subscription"`).

## Docs

The description says both "Subscribes to the device's `0x89` flow-sample frame" and "Flow samples arrive
as `0x0b` frames" — `0x89` is the subscribe request; samples come back as type `0x8B` / seq `0x0B`. The
docstring in the test file has it right.

Also worth noting in the README that only HA-initiated runs are measured — a run started from the timer's
physical button or the B-hyve app won't trigger the subscription. (The `start_watering` service is fine,
it routes through `async_open_valve`.)

ljmerza:
**Correction on the test count.** My "202 doesn't reproduce" note was wrong — my local `main` was two commits stale and missing the 5 tests #42 added. Re-run against this PR's merge ref: **202 passed** under the declared test deps, exactly as you said. Apologies for the doubt.

The two errors in blocking item 1 do still stand, but only when `pytest-homeassistant-custom-component` is installed — its autouse `verify_cleanup` fixture calls `time.monotonic()` at teardown and hits the exhausted iterator. With just `tests/requirements-test.txt` nothing trips over it, which is why your run was clean. `validate.yaml` now runs both configurations (#44), so CI will show this directly on your next push.

Nothing else in the review changes.

FabsFabios:
Hi @ljmerza ,

Thank you for taking the time to review my PR.

Status: Captures from four fw0085 units settle item 5 and turned up a blocking issue that changes the shape of item 6, so I'm posting findings before writing the rework rather than after.

## Item 5: The guard, and why half of it would break the feature

`pt[2]` is not a type byte. It's a push message counter cycling 0x80–0xBF, shared across flow, status and battery pushes; It wraps 191→128 twice in a 3-minute capture, and apparent gaps in the flow sequence are the battery and run-end frames taking their turn. The `0x8B` in my test fixture was one frame's counter value that I mistook for a type; since the production guard never reads `pt[2]`, nothing ever contradicted it. The `0x40` bit is clear across that entire range, so the TX-echo guard would have rejected 100% of flow frames permanently, not intermittently.

Taking from item 5: `SEQ_FLOW = 0x0B` as a named constant, and the `pt[4] == D747_ROUTING` check. Not taking: the type-byte check and the TX-echo guard.

## Payload layout: Three fields, not four pad bytes

`[mesh:2][push_ctr:1][seq:1][routing:1][rate u16][elapsed u16][cumulative u16]`

- `pt[7:9]` elapsed seconds since START. +3 on every one of 156 in-run samples across two devices, no exceptions. Device-side clock.
- `pt[5:7]` integer ticks/second over the preceding window. Too coarse to use: at 118 counts/L each step is 0.51 L/min.
- `pt[9:11]` cumulative ticks, reset per run.

Run end is a seq `0x0C` frame carrying the run duration at `pt[6:8]` LE - 0x00B4=180 on two 3-minute runs, 0x003C=60 on a 1-minute run.

## Items 1 and 4: Resolved by the device's own fields

Reading `dt` from the elapsed field removes `time.monotonic()` from the decode path entirely. There is then no clock for a test to patch, so the `verify_cleanup` teardown failure has nothing left to trip over. Item 1 dissolves rather than being worked around, and the two tests get rewritten to feed elapsed values in the frames.

For item 4: after the run ends the device keeps sending zeroed samples for as long as the link is open (35 and 42 of them in the two captures, still going when I stopped). Rate reaches zero from device data with no `inc == 0` branch. The trap is that cumulative also drops to 0, so computing volume directly would wipe every run total. `elapsed == 0` is the discriminator.

Reading cumulative directly instead of accumulating deltas also fixes an undercount I did not spot: the first sample of every run is currently discarded as a baseline, but the counter resets per run, so those ticks are real water. 115 ticks (1.0 L) and 47 ticks (0.4 L) in these two runs.

## Item 6: Confirmed as a race, but the suggested fix won't work

Flow samples stop after exactly 60 seconds. `_arm_idle_timer()` is called only from write paths (connection.py:401, 431, 478); inbound notifications never reset it, so the link closes 60 s after the subscribe write regardless of traffic. At the default `idle_disconnect` a 10-minute program records roughly 28% of its water. Raising the option to 600 gave complete coverage,  93 and 99 samples against 20 before, both runs covered end to end. The device honours the 700-sample request; there is no device-side cap.

That means flow needs the session held for the run duration, which collides with `start_watering`'s unconditional `finally: await self.connection.disconnect()`. Sending the subscribe inside that block as suggested would tear the link down
immediately and yield no samples at all. It also collides with the ephemeral connect→command→disconnect model, and holding the session locks out the B-hyve app and any status poll for the whole program.

An earlier iteration of this branch had a 60 s keepalive that re-sent the subscribe frame. I removed it in cleanup, reasoning that a keepalive resets a timeout and a sample budget has nothing to reset. Right about the subscription, wrong about the BLE link it was incidentally holding open.

Four 10-minute calibration runs on a build predating that removal recorded complete volumes (185 / 220.61 / 129.09 / 31.53 L, each consistent with a full program at the tick rates measured this week). I can't prove the keepalive was the mechanism, but it's the only difference I can identify, and a periodic write is the one thing that demonstrably resets `_arm_idle_timer`.

Three shapes I can see, and this is your call:

1. Run-scoped session: suppress the disconnect when flow is enabled, explicit unsubscribe on STOP. Cleanest, but holds the device's only BLE slot.
2. Periodic keepalive write for the run duration: A write, not a reconnect, so cheaper than it sounds on AA cells, but still ~10 extra writes per 10-minute program.
3. Flow opt-in per device, so the session cost is only paid where wanted.

## Items 2 and 3: Accepted as filed

`connection.py` reverts entirely; I've confirmed the duplicate
`set_plaintext_observer` is exactly the four added lines and the observer routing already exists via `base.py:310`. `_observe_plaintext`'s docstring restored, with the call moved below `super()._observe_plaintext(pt)` to match existing ordering.

## Non-blocking: All accepted

`has_flow_gen1` and `flow_counts_per_litre` declared on `BHyveBleDeviceBase` following the `has_flow` pattern, plain attribute access, both added to diagnostics. `_loop()` goes away with whichever session design you pick.

Gen1 Water used extends `_RestoreLastValueSensor` and reports 0.0 rather than None. 

`number.py` setup skips `connection is None` matching `sensor.py:78`.

Function-local `import time` and `import asyncio` both disappear with the above. Clamp literals become
`MIN_/MAX_COUNTS_PER_LITRE`. 

`valve.py` reverts to upstream; With the subscribe moved into the device class it needs no change at all, so the diff drops from 7 files to 5.

## Docs

Description corrected: `0x89`/`0x0E` is the subscribe request, samples come back on seq `0x0B`, run end on seq `0x0C`, and there is no fixed type byte.

README will note that only HA-initiated runs are measured, and, depending on your answer will be to above, what the flow coverage limit is.

ljmerza:
Excellent write-up — the captures answer more than they were asked to, and two of them correct me.

## Item 5: you're right, and my guard would have broken it

`pt[2]` as a push counter cycling `0x80–0xBF` explains everything, including why nothing ever contradicted the production guard. My suggested `pt[2] == 0x8B` check was reading one frame's counter value as a type, and the TX-echo guard is worse than that — with the `0x40` bit clear across the whole range it would have rejected 100% of flow frames, permanently. Dropping both is correct. `SEQ_FLOW = 0x0B` plus `pt[4] == D747_ROUTING` is the right guard.

Same for item 6: sending the subscribe inside `start_watering`'s locked block hits the `finally: disconnect()` and yields nothing. That suggestion was wrong too.

## So how does it work today? It doesn't

Worth stating plainly, because the PR description doesn't: at default settings this feature captures the first ~60 seconds of a run and nothing after. Your 28% on a 10-minute program, and the 20-vs-93/99 sample counts, are the real numbers. The 3000 ms × 700 budget is sized for 35 minutes of a link that is gone at minute one.

Your four complete 10-minute calibration runs are the outlier that needs explaining, and I agree the keepalive is the likely mechanism — a periodic write is the only thing that demonstrably re-arms `_arm_idle_timer`. **What is your `idle_disconnect_sec` set to?** If it's also been raised from 60, that's a second variable in those runs and worth separating before we design around them.

## The three shapes — 3 isn't a peer of 1 and 2

Option 3 is the gate, not a third session design. It composes with either of the others, and I think it's required regardless of which you pick. Taking 1 and 2 on their own:

**Option 1 needs more than removing the disconnect.** Suppressing `start_watering`'s `finally` doesn't stop `_idle_close()` — `_arm_idle_timer` is armed by the subscribe write and fires 60 s later no matter what the link is doing. So option 1 implies a real connection-level concept: something like a counted `hold_open()` / `release()` on `BHyveBleConnection` that disarms the idle timer while held. That's more code, but it makes the intent explicit and testable instead of an emergent property of whatever `idle_disconnect_sec` happens to be — which is the actual defect here.

**Option 2 works with existing machinery, but isn't cheaper.** Each keepalive write re-arms the timer; that's precisely why your pre-cleanup build recorded complete volumes. But it doesn't avoid holding the session — it holds the same session and adds ~10 writes on top. On battery it's option 1 plus traffic, not an alternative to it.

So 1 and 2 are the same order of cost, and 1 is the cleaner of the two.

## Battery: yes, and it can't be on by default

This repo has already priced this exact tradeoff twice, in its own comments:

- `ht25.py`: *"Never while watering: at the watering cadence a connect per tick would hammer the AA cells and race actuations."*
- `const.py`: mesh status poll defaults **off** — *"~96 connects/day at the default 15-min idle cadence, each a full bind/init on AA cells."*

We already decided ~96 short connects a day was too expensive to enable by default. A sustained link for the full duration of every run is more radio-on time than that. Whichever session design wins, it can't be the default.

And there's the non-battery cost you already identified: holding the device's only BLE slot locks out the B-hyve app and any status poll for the whole program. On a shared setup that's user-visible in a way battery drain isn't.

## My call: option 1, gated behind option 3, default off

- **Per-device opt-in**, following `BHyveBatteryChemistrySelect` in `select.py:137` — `CoordinatorEntity` + `RestoreEntity` + `EntityCategory.CONFIG`, `unique_id = f"{device.unique_id}_..."`. A config switch is the natural shape.
- **Off means off**: don't send the subscribe at all, and don't create the flow entities. No session cost, no dead sensors on devices that aren't measuring.
- **Prefer 1 over 2** — a held link with no traffic beats a held link plus periodic writes.
- **Release on every exit path**, not just `async_close_valve`: the wall-clock auto-close in `coordinator._async_update_data`, and `stop_all` (`__init__.py:341`) which calls `device.stop_watering()` directly and bypasses the valve entity entirely. A run that ends by timeout is the common case, and that's the path that most needs to drop the link.
- Document the cost on the option itself, the way `DEFAULT_MESH_STATUS_POLL` does.

## One regression in the new decode plan

Reading `cumulative` directly instead of accumulating deltas loses wrap handling, and the wrap is reachable.

At 118 counts/L, `pt[9:11]` wraps at 65536 ticks = **555 L**. Your own calibration runs recorded 220.61 L in 10 minutes — about 22 L/min. A 30-minute program at that rate is ~660 L, past the wrap, and a direct read would report ~105 L.

Suggest keeping a wrap counter: cumulative decreasing *while elapsed increases* is a wrap; `elapsed == 0` still discriminates the post-run zeroed samples. That keeps both the undercount fix and wrap safety.

## Everything else

The elapsed-field change is a clear improvement — pulling `time.monotonic()` out of the decode path dissolves item 1 instead of working around it, and the first-sample undercount is a good catch I'd missed entirely. Run-end on seq `0x0C` with the duration at `pt[6:8]` is worth a decode too, if only to end the run cleanly rather than inferring it from zeroed samples.

Items 2, 3 and the whole non-blocking list sound right as you've described them. Confirmed on my side that `connection.py` reverts cleanly — `base.py:308` already wires `set_plaintext_observer(self._observe_plaintext)`, so those four lines were pure duplication.

Take the session decision as settled from my side: **run-scoped hold, per-device opt-in, default off.** If you'd rather do the keepalive because it's less connection-layer surgery, I'd take that too as long as it's behind the same opt-in — but I don't want either shape on by default.

FabsFabios:
`idle_disconnect_sec` was 60 throughout — provable rather than recalled. The two truncated captures cut off at 60.008 s and 60.009 s, the default to the millisecond. It was raised to 600 only for the controlled comparison and reverted after. So there's no second variable in the calibration runs: the keepalive is the only identified difference.

Session design accepted as settled — run-scoped hold, per-device opt-in, default off. Agreed that 3 is a gate rather than a peer, and that option 2 is option 1 plus traffic rather than an alternative to it. Agreed too that suppressing the `finally` isn't sufficient, since `_arm_idle_timer` is armed by the subscribe write itself; `hold_open()`/`release()` on the connection is the
right shape.

The wrap catch is correct and I should have caught it. At Driveway Right's 40-41 ticks/s the u16 wraps at ~27 minutes — before the 700-sample budget expires at 35, so it's inside the design envelope. Keeping a wrap counter, with `elapsed` discriminating: never 0 during a run (first samples read 8 and 9), always 0 in the post-run zeroed frames.

Three things I'd rather settle now than build twice:

1. **Entity lifecycle.** Correcting my own premise first: I assumed changing a config entry option reloads the entry. It doesn't. `_async_update_listener` (`__init__.py:162`) deliberately applies options in place because a reload rebuilds every `DeviceState` — wiping an in-progress run — and immediately reconnects, which the single-BLE-session constraint turns into broken comms until a full HA restart. So `DEFAULT_MESH_STATUS_POLL` doesn't dodge the entity problem by reloading; it dodges it by creating no entities at all and just flipping `active_status_poll` on the device object.That means nothing — option or switch — can create entities after `async_setup_entry`. So "off means don't create the flow entities" isn't achievable as stated, and the real question is what an off device shows.

Two shapes:

**(a) Create the flow entities whenever `has_flow_gen1` and the record has a connection; gate only the subscribe and the session hold on the switch.**
   Report `unavailable` while the switch is off, via the `available` property — which is what `BHyveProgramEnableSwitch.available` (`switch.py`) already does for program slots with nothing stored in them. Same treatment for `BHyveFlowCalibrationNumber`: nothing to calibrate while nothing is measuring.

**(b) Stash `async_add_entities` in runtime data and add the entities on first enable.**

I'd take (a), and not only for the smaller diff. (b) is exact only until the first time someone enables the switch — entities can't be withdrawn on disable without orphaning their recorder history, so from that point on (b) behaves identically to (a) forever. It buys precision for the never-enabled-once case only, and (a) already covers that case with `unavailable`, which is a truthful state rather than a dead sensor: the   device genuinely isn't measuring.

(a) also keeps entity management out of the switch entirely, so the switch stays a behaviour gate — the same shape as mesh status poll, one layer down. Happy to build (b) if you'd rather it be exact for that first case.

2. **Hold expiry.** If all four release paths are missed the link stays open indefinitely, which is the failure the opt-in exists to prevent. Suggest `hold_open()` taking a max duration and self-releasing, so a leaked hold costs one run rather than the batteries.

3. **Sample budget.** Still hardcoded at 3000 ms x 700 = 35 min while `MAX_MINUTES` allows 1440. `start_watering` knows `duration_sec`, so I'd size the count from it with a margin. Do you know of a device-side ceiling on that field? 700 was honoured; I haven't probed higher.

Also taking the seq `0x0C` decode — under this design it's the natural release trigger and more reliable than the wall-clock backstop, though it needs scheduling rather than releasing inline since it arrives on the notification callback.

Two additions to the release list:

- `async_unload_entry` (`__init__.py:154`) calls `device.async_unload()`, which disconnects (`base.py:351`). A hold has to be released there too, or unload races whatever is holding the link open. That's the path that fires on HA shutdown and on integration reload.
- `_async_update_listener` writes `connection._idle_sec` in place (`__init__.py:191`). That write must not re-arm a timer an active hold has disarmed.

Also noting `switch.py` currently skips non-protobuf devices entirely, so this adds the first HT25 branch to that platform rather than extending an existing list.

Will hold off writing until you've weighed in on 1-3.

ljmerza:
The 60.008 s / 60.009 s cutoffs are the kind of answer I was hoping for — that closes the calibration-run question completely. And the wrap being reachable at ~27 minutes on Driveway Right, inside the 35-minute budget, upgrades it from theoretical to design-envelope; good.

You also caught a flaw in my "off means off": I'd assumed entity creation could follow the option, and the no-reload constraint in `_async_update_listener` makes that unachievable as stated. Verified your reading on my side — including `async_unload_entry` → `async_unload()` → disconnect, and the in-place `_idle_sec` write. Both of your release-list additions are correct, and the second one settles an architectural point: hold state has to live on the connection itself, where the idle-timer paths and the `_idle_sec` setter can respect it — which is another argument for `hold_open()`/`release()` being a `BHyveBleConnection` concept rather than device-class bookkeeping.

Answers to 1–3:

## 1. Entity lifecycle: (a)

Agreed, and for your reasons — (b) is exact only until the first enable, then behaves as (a) forever, so it buys precision for one case that `unavailable` already handles truthfully. `BHyveProgramEnableSwitch.available` is the right precedent to mirror.

One amendment: apply `available` gating to the two **sensors** only, not to `BHyveFlowCalibrationNumber`. The calibration number is `EntityCategory.CONFIG` describing a property of the plumbing, not a live measurement — keeping it settable while the switch is off lets someone dial in a known counts/L before their first measured run, and a RestoreNumber that's unavailable-but-holding-state is an awkward combination. There's nothing it can break while nothing is measuring.

## 2. Hold expiry: yes, and derive the cap

`start_watering` knows `duration_sec`, so make it `hold_open(max_sec=duration_sec + grace)` rather than inventing a constant — `EXPIRY_GRACE_SEC` is the existing precedent for the margin. Two shapes worth pinning down:

- Self-expiry should **re-arm the idle timer, not disconnect directly**. Release means "hand the link back to the normal machinery"; the existing 60 s idle path then closes it. That keeps exactly one place in the code that actually tears down sessions.
- Make `hold_open()`/`release()` counted or idempotent, whichever is cheaper — with five release paths (STOP ack via seq `0x0C`, `async_close_valve`, wall-clock auto-close, `stop_all`, unload), double-release must be harmless.

And agreed on the `0x0C` decode as the natural release trigger, scheduled off the notification callback rather than inline — `hass.add_job` is the existing pattern for exactly that.

## 3. Sample budget: derive it; no known ceiling

I don't know of a device-side ceiling, and nothing in any capture so far suggests one — the only hard bound is the u16 field itself (65535). 700 is simply the highest value verified. So:

- `count = ceil(duration_sec * 1000 / interval_ms) + margin`, clamped to 65535.
- You have four units — probe a large count (a few thousand) on one of them before relying on it; that's cheaper than designing around a ceiling that may not exist.
- For pathological durations (`MAX_MINUTES` allows 1440), stretch the **interval** rather than the count. The wrap counter only needs sampling comfortably faster than the fastest observed wrap period (~27 min), so even a 60 s interval keeps volume exact; only rate resolution degrades, and a 24-hour soak doesn't need 3 s rate resolution.

On `switch.py` gaining its first HT25 branch: fine — gate on `has_flow_gen1` like the other platforms do, rather than device-class isinstance, and it stays one line.

Nothing else outstanding from my side. Looking forward to the rework — with the CI from #44 in place, the plugin-fixture class of failure will show up in the PR checks directly this time.

FabsFabios:
Three probes since your last message, and they retire the budget guidance entirely — including a claim of mine that your answer was built on.

## Correction: the count field is ignored

I told you "the device honours the 700-sample request; there is no device-side cap." That was wrong, and I said it on the strength of captures that never exceeded 99 samples. Two bounded probes:

| interval | count requested | samples | span | ended by |
|---|---|---|---|---|
| 3000 ms | 700 | 97 | 291 s | device, idle_disconnect at 3600 |
| 1000 ms | 700 | 297 | 296 s | device, idle_disconnect at 3600 |

Tripling the rate tripled the sample count and left the duration unchanged. The interval field is honoured exactly (1.001 s spacing, no gaps); the count field is ignored. The device samples for ~300 s from the subscribe and stops. No link
drop logged in either case. That kills both budget approaches. Deriving the count does nothing, probing a few thousand would have measured nothing, and stretching the interval for long runs is strictly worse — a 10-minute program at 6000 ms still stops at 300 s, just at half the resolution.

## Re-subscription renews the budget — confirmed on hardware

6-minute run, re-subscribe every 240 s, idle_disconnect raised for the test:

- 412 flow samples over 20.6 minutes, zero gaps, six renewals
- the complete run captured: 118 in-run samples, elapsed 8 to 359
- 11,947 ticks at elapsed 359 / 114 counts/L = 104.8 L over 360 s = 17.5 L/min, matching the vendor app's figure for this timer exactly

So option 2 isn't option 1 plus traffic — it's the only shape that covers a program longer than five minutes, with option 1's hold as the substrate under it. This also explains the complete 10-minute calibration runs: the keepalive I removed in cleanup was re-sending the subscribe frame every 60 s, renewing the 300 s budget as well as re-arming `_arm_idle_timer`. My PR notes claim "re-subscribing mid-run is ACKed but yields no further samples" — that observation came from a session that had already disconnected underneath it. Retracted.

## Correction: no re-anchoring

I said a mid-run re-subscribe re-anchors the counters and the decode needs continuation logic. It doesn't. Across a mid-run subscribe:

- before: elapsed 245, cumulative 8121
- subscribe sent
- after: elapsed 248, cumulative 8222

Seamless. Both fields are anchored to START, not to the subscription. My earlier reading came from a first sample of `elapsed=6, cum=0`, which was ramp-up before water reached the meter — the next sample read `elapsed=7, cum=44`. So `elapsed` decreasing means a new run and nothing else. One branch less.

## Correction: the run-end frame's second field is not a total

I was about to tell you it carried the authoritative run total. Across six captures on three devices, `pt[9:11]` on the seq 0x0C frame is exactly `duration x integer`:

| capture | duration | pt[9:11] | ratio |
|---|---|---|---|
| DL 6 min | 360 s | 12,240 | 34.0000 |
| DL 1 min (x2) | 60 s | 2,040 | 34.0000 |
| DR 3 min | 180 s | 7,920 | 44.0000 |
| LP 3 min | 180 s | 900 | 5.0000 |
| LP 1 min | 60 s | 360 | 6.0000 |

On the run where I can check precisely it reports 12,240 against a measured 11,947 at elapsed 359 (so ~11,980 at 360) — about 2.2% high. The multiplier matches neither the average, steady, nor final rate on any device. Unidentified, and not usable as a volume source. Only `pt[6:8]` (run duration) is established on that frame, and its value is as a release trigger.

## Two lifecycle failures, both observed rather than hypothesised

**A naturally ending run releases nothing.** The 6-minute run ended and zeroed samples kept arriving for ~15 minutes, because a run that ends on the device's own timer never calls `async_close_valve`, so nothing cancelled the keepalive.
Exactly the case you flagged — release on the wall-clock auto-close path isn't defensive, it's the common one.

**The task survives config entry unload.** I reloaded the integration to stop the transmission. It didn't stop: three more re-subscribe cycles fired at 15:08:01, 15:12:08 and 15:16:13, each reconnecting and re-running the 8-step init on a device the current integration instance no longer knew about. Only a full HA restart ended it.

`hass.async_create_task` isn't tied to the config entry, and `async_unload_entry` -> `async_unload()` only disconnects (base.py:351) — `stop_flow_subscription` is reachable solely from `async_close_valve`. That is present in the PR as filed. It's benign there only because the send-once task completes immediately; with a keepalive loop it becomes a permanent BLE reconnect every four minutes on AA cells, surviving reload.

Proposed fix: an `async_unload()` override on BHyveHT25Device that cancels `_flow_sub_task` before calling super(), rather than `entry.async_create_background_task()` — the latter would put ConfigEntry knowledge into the device layer, which nothing in devices/ currently has.

Incidentally that accident re-confirmed the idle timer independently: with idle_disconnect back at its default, each of the three orphaned subscriptions ran 60.006, 60.008 and 60.008 s from its subscribe ack before the link closed.

## Where the design lands

- `hold_open()` / `release()` on BHyveBleConnection: counted, self-expiring on `duration_sec + EXPIRY_GRACE_SEC`, re-arming the idle timer rather than disconnecting, and respected by the `_idle_sec` setter.
- Re-subscribe every 240 s while the valve is open (300 s budget, 60 s margin). About two extra writes on a 10-minute program.
- Both stop at six paths, not five: seq 0x0C, `async_close_valve`, wall-clock auto-close, `stop_all`, `async_unload_entry`, and task cancellation on device unload.
- Volume from the sample stream's cumulative field with a wrap counter — still needed, the u16 wraps around 27-32 minutes at these flow rates — not from the run-end frame.
- Per-device opt-in switch, default off, entities created whenever `has_flow_gen1` and gated via `available`; calibration number left settable as you specified.
- Subscribe stays 3000 ms / 700. The count is ignored, so there is nothing to derive, and the interval stays put because stretching it buys nothing.

Protocol work is finished: interval honoured, count ignored, 300 s budget renewable at 240 s, no re-anchoring, run-end second field unidentified, task lifecycle understood. Building this now unless you want a different shape.

ljmerza:
Three probes that each overturned something — including my entire budget answer, which you've saved me from: derive-the-count and stretch-the-interval were both built on the count meaning something, and it doesn't. The renewal result is the important one; with a gapless 20.6-minute capture and the decoded total matching the vendor app exactly, the keepalive goes from workaround to load-bearing, and my "option 2 is option 1 plus traffic" framing inverts with it. The reload-survival find is the best catch in the thread — an orphaned task re-running the 8-step init every four minutes against a device the integration no longer owns is precisely the failure class the opt-in exists to prevent, and your `async_unload()` override is the right fix for the right layering reason.

**Build it.** The design as you've landed it — counted self-expiring hold on the connection, 240 s renewal, six stop paths, cumulative + wrap counter, opt-in default off, entities gated `available`, calibration settable — is what I'd merge.

Two notes to fold in, neither a redesign:

1. **"Ignored" may be "capped at 300 s."** Both probes ran count=700, where count × interval far exceeds 300 s — they can't distinguish a device that ignores the field from one that clamps it. The vendor app's captured reference asked for 30 × 1000 ms = 30 s, which is a strange thing to request if the field does nothing. One small-count probe (10 × 1000 ms: stops at 10 s or 300?) settles it. No design impact either way — 3000/700 exceeds the cap under either reading and the renewal covers the rest — but the README/protocol notes should say which it is rather than guessing.

2. **The renewal write needs `_api_lock`.** The original interleaving concern now applies to a write that fires repeatedly *during* watering — concurrent with any STOP actuation and the watering-cadence poll on the device's single session. Take the lock around each re-subscribe send (and skip the cycle rather than queue behind a long-held lock, so a renewal can't pile up behind an actuation and fire late into a closed session).

And one test to make sure exists, since it's now the scariest path: reload the config entry mid-run and assert the keepalive task is cancelled — the regression test for the failure you watched happen.

FabsFabios:
Pushed. The small-count probe is in, and it goes your way.

`10 x 1000 ms` gave exactly 10 samples, stopping 45 s into a 60 s run. So both fields are honoured and there's a ~300 s ceiling on top: the device stops after `count` samples or ~300 s, whichever comes first. My "the count is ignored" was wrong — 700 simply always loses to the ceiling. Every observation fits `min(count, 300s / interval)`: 10, 97, 297.

No design change. 700 stays, deliberately large enough that the ceiling always binds, so stream length is predictable whatever the interval and renewal covers the rest. `_flow_subscribe_frame` now states it as measured fact with all three data points rather than a guess.

One bonus: the seq `0x0C` run-end frame arrived 46 s **after** that 10-sample subscription had expired, so it isn't part of the flow budget. That makes it reliable as a release trigger even when sampling has lapsed mid-run.

Everything from the review:

- 1 — `time.monotonic` is gone from the decode entirely, so there's no clock left to patch and the `verify_cleanup` teardown failure is structurally impossible rather than worked around
- 2 — `connection.py` reverted to upstream
- 3 — `_observe_plaintext` docstring restored, call moved below `super()`
- 4 — rate zeroes on a static meter and on the zeroed post-run samples
- 5 — `SEQ_FLOW` + routing check; no type or TX-echo guard, per the push-counter finding
- 6 — subscribe inside `_api_lock` on the START session, `finally` guarded on `is_held`
- `_loop()` gone with the background task; `has_flow_gen1` / `flow_counts_per_litre` / `flow_gen1_enabled` declared on the device base with plain attribute access and added to diagnostics; Water used extends `_RestoreLastValueSensor` and reports 0.0 rather than None; `number.py` skips key-less records; function-local imports gone; clamp bounds named; `valve.py` reverted, so the diff is 10 files rather than the original 7 + churn
- Renewal takes `_api_lock` and skips rather than queues, as you specified
- Reload-mid-run regression test added and mutation-verified

Docs: description rewritten (the `0x89` / `0x8B` conflation is gone — `0x89` is the subscribe, samples come back on seq `0x0B`, run end on `0x0C`, and there is no fixed type byte). README documents the opt-in and its cost, the calibration
method with the 113–118 range, and the HA-initiated-runs-only limitation. The options-flow "Flow calibration (counts per gallon)" entry now says explicitly that it does nothing on Gen1.

PR #45, opened 2026-07-28
ci: make the HA test-plugin pin visible to Dependabot

ljmerza:
Follow-up to #44. The `pytest-homeassistant-custom-component==0.13.205` pin sat in a workflow input string, where Dependabot can't see it. That plugin pins an exact Home Assistant version, so with no bumps the nightly cron quietly tests an ever-older HA.

## Changes

- **`tests/requirements-ha.txt`** (new) — holds the pin, with a comment explaining what the plugin buys us and why it's a separate file.
- **`validate.yaml`** — `extra-packages: -r tests/requirements-ha.txt` instead of the inline pin. The action forwards `extra-packages` to `uv pip install` unquoted, so `-r <file>` passes straight through.
- **`.github/dependabot.yml`** (new) — `pip` on `/tests` (picks up both requirements files) and `github-actions` on `/`, both monthly.

`tests/requirements-test.txt` is deliberately untouched: the suite still needs no Home Assistant, and the `Tests` job keeps proving that.

## Verified

Ran the action's steps end to end on Python 3.13 against this tree:

```
install exit=0
plugin installed OK
186 passed in 32.15s
```

So the `-r` indirection installs the plugin exactly as the inline pin did.

## Note

Dependabot can't do anything about `ljmerza/misc-actions/actions/python-test@main` — only tagged or SHA refs are actionable. Called out in a comment in the config so it isn't mistaken for an oversight.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+version: 2
+
+updates:
+  # Action refs in .github/workflows/. Note validate.yaml tracks the shared
+  # python-test action at @main, which Dependabot cannot pin or bump — only
+  # tagged/SHA refs are actionable here.
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+    commit-message:
+      prefix: ci
+
+  # tests/requirements-test.txt and tests/requirements-ha.txt. The latter holds
+  # the pytest-homeassistant-custom-component pin, which drags in an exact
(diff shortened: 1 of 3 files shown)


ljmerza:
Closing — Dependabot isn't wanted here, and this change existed only to make the pin visible to it. The inline pin in validate.yaml stays as merged in #44.

PR #46, opened 2026-07-31
fix: replace hardcoded network-key table with a user option

ljmerza:
`devices/ht25.py` carried `_HUB_MESH_BY_NETWORK_KEY`, a dict mapping two 32-hex strings to hub mesh ids. Those strings are AES-128 BLE network keys — `connection.py` does `bytes.fromhex(network_key)` and feeds the result to `AES(self._key)` — so the repo was publishing working keys for two real mesh networks. `tests/test_diagnostics.py` carried one of them as a fixture.

Reported privately by @FabsFabios, who found it during a static pass and deliberately left it out of #43 rather than ripping unrelated working code out of someone else's project. Good call, and the right way to handle it.

## Why not just delete the dict

The table is a live fallback: `hub_mesh_address` uses it when the cloud record has no `bridge_device_id`, and `magic2` needs a hub id. Deleting it alone `NameError`s, and deleting the lookup too would silently drop affected setups onto the `0x0000` placeholder — the existing comment already warns that's a suspect when START gets dropped.

So the value moves out of source rather than disappearing:

- **New `hub_mesh_overrides` option** — comma-separated `MAC=id` pairs, id decimal or `0x`-hex, e.g. `AA:BB:CC:DD:EE:FF=0xEB42`. Keyed by **MAC**, so configuring it needs nothing secret.
- `parse_hub_mesh_overrides()` is a pure function, HA-free and unit-tested, following the `accumulate_gallons` precedent. Malformed pairs are skipped with a warning rather than failing setup — one typo shouldn't take the integration down.
- Applied in place by `_async_update_listener` alongside the other tuning options, so an edit takes effect without the reload that would wipe in-progress run state.
- `hub_mesh_address` resolution is now: cloud record → user override → `0x0000` placeholder + warning.

## Impact

Unchanged for anyone whose cloud record has `bridge_device_id`. Anyone relying on the removed table lands on the placeholder path and can restore exact prior behaviour by setting the option — the warning names the option so it's self-directing.

## Verification

- 194 passed (186 baseline + 8 new).
- New `tests/test_hub_mesh_override.py` covers the parse format, MAC upper-casing, malformed and out-of-range tolerance, and all three resolution tiers.
- Scripted check that neither key appears anywhere in the working tree — clean.
- Test fixture replaced with an obvious dummy (`00112233...eeff`).

## What this does NOT fix

Both keys remain in git history from the initial commit, and in every existing clone and fork. **They should be treated as compromised regardless of this PR.** The exposure is proximity-gated — an attacker needs BLE range — but within range the keys allow decrypting and forging valve traffic. Rotation (if the devices support it) is the real remedy; history rewriting is a separate decision and can't reach existing forks.

### DIFF
--- a/custom_components/orbit_bhyve/__init__.py
+++ b/custom_components/orbit_bhyve/__init__.py
@@ -33,6 +33,7 @@ from .const import (
     CONF_EMAIL,
     CONF_FLOW_COUNTS_PER_GALLON,
     CONF_IDLE_DISCONNECT,
+    CONF_HUB_MESH_OVERRIDES,
     CONF_MESH_STATUS_POLL,
     CONF_PASSWORD,
     CONF_POLL_IDLE,
@@ -40,6 +41,7 @@ from .const import (
     DEFAULT_DURATION,
     DEFAULT_FLOW_COUNTS_PER_GALLON,
     DEFAULT_IDLE_DISCONNECT,
+    DEFAULT_HUB_MESH_OVERRIDES,
     DEFAULT_MESH_STATUS_POLL,
     DEFAULT_POLL_IDLE,
     DEFAULT_POLL_WATERING,
@@ -47,6 +49,7 @@ from .const import (
 )
 from .coordinator import BHyveDeviceCoordinator
 from .devices import BHyveHT25Device, UnsupportedModel, build_device
+from .devices.ht25 import parse_hub_mesh_overrides
 from .devices.base import (
     PROGRAM_SLOTS,
     SLOT_LETTERS,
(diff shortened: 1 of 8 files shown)


FabsFabios:
Thanks for turning this round so fast, and for the MAC-keyed option. Keying on something non-secret is the right shape, and it means anyone who was relying on the table can restore exact behaviour without needing the key to do it.

One thing worth a look, pre-existing rather than introduced here but this PR makes it matter more.

`hub_mesh_address` is evaluated twice per post-handshake init. Once in the debug log at ht25.py:162 (the argument still evaluates with debug off) and once building the magic2 payload at :173. Init re-runs on every actuation, so a
device on the placeholder path emits that warning twice per watering command, indefinitely.

That was nearly invisible while the table caught the affected setups. With it gone, everyone whose cloud record lacks `bridge_device_id` lands there. A once-per-device latch, or dropping it to debug after the first emission, would keep it self-directing without becoming background noise. Happy to send it as a follow-up if you'd rather keep this PR focused.

On rotation: I don't know whether the HT25 supports it either. The integration reads the key from the cloud record, and I haven't tested whether re-pairing or a factory reset regenerates one. Worth establishing before deciding how much effort a history rewrite is worth. If the keys are fixed per unit, removal from source is the only remedy available and the calculus changes.

I'll rebase #43 once this merges.

PR #48, opened 2026-08-07
fix(devices): key HT25 mesh vs protobuf routing on hardware suffix, not firmware

### LINKED ISSUE #47: New HT25, came with FW0098. Any chance this might be able to be added in the future?
Is there any way to downgrade the FW to the supported version?

The integration pulls the timer into HA, but it never connects to the BTProxy in the garage, which I have about 10ft away. The HT34 I have out back is working flawlessly to another BTProxy on the other side of the wall in the backyard!!

Thanks for pulling this together!

exactmike:
Fixes #47.

### The bug

`resolve_device_class()` disambiguates the shared `"HT25"` hardware prefix between the
mesh (d7-47) and protobuf (0x11) protocol families by checking
`hardware.startswith("HT25G2")` OR `firmware == "0111"`, and falls back to the **mesh**
class for anything matching neither.

That default is backwards. Every mesh device ever observed reports hardware
`HT25-0000`, while the known non-`G2` protobuf hardware string (`HT25-0001`) has no such
suffix. Pinning the protobuf branch to the single firmware literal `"0111"` means any
newer firmware build on protobuf-family hardware that isn't `G2`-prefixed silently
misroutes to the mesh class, which speaks a different wire protocol. The device still
gets discovered and added (cloud-backed setup doesn't touch the wire protocol), but the
BLE session never completes and the device stays unusable.

This is exactly what #47 reports: a `90205Z` (Gen2 single-station) unit on `fw0098`
reporting hardware `HT25A-0001` — a variant not covered by either branch of the old
check — gets added but never connects.

### The fix

Check the explicit `HT25G2` prefix first, then key the remaining fallback off the
`-0000` hardware suffix (the only signal ever observed on real mesh hardware) to select
the mesh class; everything else under the `HT25` prefix defaults to the protobuf class.
This removes the firmware whitelist entirely, so future firmware drift on either family
can no longer misroute a device across protocol families.

Routing for `HT25-0000` (mesh) and `HT25G2-0001` (protobuf) on every previously-observed
firmware is unchanged. One combination does change: `HT25-0001` on `fw0085`/`fw0041`
routed to mesh under the old firmware whitelist; those rows in the prior test suite were
synthetic (no unit reporting that hardware/firmware pair has actually been observed), and
the rest of the codebase already treats `HT25-0001` as protobuf-family hardware. This PR
routes it there consistently instead.

### Verification

- `resolve_device_class()` unit tests in `tests/test_devices.py` (`test_resolve_routes`),
  including the `HT25A-0001` string from #47 and a case confirming the `HT25G2` prefix
  takes priority over the `-0000` suffix.
- 231/231 local pytest green.
- Hardware-verified by @quadcom (issue #47) on his `90205Z` unit
  (hardware `HT25A-0001`, fw `0098`) — confirmed the branch resolves the device to the
  protobuf class and it connects/actuates correctly after a full HA restart (not just a
  config-entry reload — custom-component modules are cached in-process, so file changes
  need a real restart to take effect).

### Notes for review

- Updated per review: `HT25G2` prefix now checked before the `-0000` suffix, added the
  `HT25A-0001`/`HT25G2-0000` test cases, fixed the stale dispatcher comment in
  `ht25g2.py`, and added the `HT25A-0001` row to the README support table.
- Following up on #47 to ask whether the flow sensor populates on quadcom's unit, since
  `has_flow` now applies to `HT25A-0001` and `HT25-0001` as a side effect of routing —
  will report back here.
- Diff is scoped to `custom_components/orbit_bhyve/devices/__init__.py` (dispatcher),
  `devices/ht25g2.py` (comment), `tests/test_devices.py`, and the README table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ when the WAN goes down.
 | Hose-tap timer    | `HT25-0000`    | `0085`           | ✅ Actuated end-to-end                   |
 | Hose-tap timer    | `HT25-0000`    | `0041`           | ✅ Actuated end-to-end (per-device mesh-ID addressing) |
 | Hose-tap timer (Gen2) | `HT25G2-0001` | `0111`          | ✅ Actuated end-to-end (protobuf protocol) |
+| Hose-tap timer (Gen2, 90205Z) | `HT25A-0001` | `0098`  | ✅ Connect/actuate verified by community member (issue #47); flow-sensor readings not yet confirmed on this variant |
 | 4-port XD         | `HT34A-0001`   | `0107`           | ✅ Battery/status decode verified on hardware; XD actuation proven via HT32A sibling |
 | 4-port XD         | `HT34-0001`    | `0058`           | ⚠️ Shares the XD protobuf protocol; not tested here |
 | 2-port XD         | `HT32A-0001`   | `0107`           | ✅ Actuated end-to-end (shares the HT34A XD protobuf protocol) |
--- a/custom_components/orbit_bhyve/devices/__init__.py
+++ b/custom_components/orbit_bhyve/devices/__init__.py
@@ -31,20 +31,23 @@ def resolve_device_class(*, hardware: str, firmware: str, type_: str) -> type[BH
     if type_ == "bridge":
         return BHyveHubDevice
     if (hardware or "").startswith("HT25"):
-        # Gen2 HT25G2 valves (fw0111) share the "HT25" hardware prefix but
-        # speak the protobuf protocol (frame magic 0x11) like the HT34A XD,
-        # NOT the d7-47 mesh protocol of the HT25-0000 hose timers. Route
-        # them away from the mesh classes before falling through. Match on
-        # the hardware suffix or fw so HT25-0000 (fw0041/0085) is untouched.
-        if (hardware or "").startswith("HT25G2") or firmware == "0111":
-            return BHyveHT25G2Device
-        # HT25-0000 mesh (d7-47) firmwares. fw0085 keeps upstream's thin
-        # subclass (retains _rebind_sid_delta=3, community-verified); fw0041 and
-        # any other fw use the parameterized base, which builds frames from the
-        # device's own mesh_device_id (not a hardcoded identity).
-        if firmware == "0085":
-            return BHyveHT25Fw0085Device
-        return BHyveHT25Device
+        # Mesh d7-47 hose timers are the only HT25 devices ever observed
+        # reporting the "-0000" hardware suffix — treat that as the
+        # authoritative signal rather than pinning to a firmware whitelist,
+        # which drifts every time Orbit ships a new build (e.g. a fw0098
+        # HT25G2 unit, upstream issue #47). Everything else under the HT25
+        # prefix — HT25G2-0001, the bare HT25-0001 variant, or any future
+        # firmware on either — speaks the protobuf protocol (frame magic
+        # 0x11) like the HT34A XD.
+        if (hardware or "").endswith("-0000"):
+            # fw0085 keeps upstream's thin subclass (retains
+            # _rebind_sid_delta=3, community-verified); fw0041 and any other
+            # fw use the parameterized base, which builds frames from the
+            # device's own mesh_device_id (not a hardcoded identity).
+            if firmware == "0085":
+                return BHyveHT25Fw0085Device
+            return BHyveHT25Device
+        return BHyveHT25G2Device
     if (hardware or "").startswith("HT34"):
         # Both HT34A-0001 and HT34-0001 (fw0058) use the protobuf XD protocol.
         # The older HT34 sharing it is the stuartdenne fork's claim (2026-06-27),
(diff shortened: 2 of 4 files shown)


ljmerza:
Reviewed — the diagnosis is correct and the fix is minimal and well-scoped. Approving with three small follow-ups.

### Verified locally

Ran the branch in an isolated worktree: **229/229 pass**, and CI is green across all four checks. Probed the dispatcher directly to compare against `main`:

| hardware | fw | main | this PR |
|---|---|---|---|
| `HT25A-0001` | 0098 | `BHyveHT25Device` (mesh — the bug) | `BHyveHT25G2Device` ✅ |
| `HT25G2-0001` | 0111 | protobuf | protobuf (unchanged) |
| `HT25-0000` | 0085 / 0041 | mesh | mesh (unchanged) |
| `HT25-0001` | 0085 / 0041 | **mesh** | **protobuf** (changed) |
| `HT25G2-0000` | 0111 | protobuf | **mesh** |

The core reasoning holds: `-0000` is the only signal ever actually observed on mesh hardware (the README rows, `test_hub_mesh_override.py` and `test_mesh_status.py` all use `HT25-0000`), and a firmware whitelist is guaranteed to drift. Misrouting is non-destructive in both directions — wrong-family frames are ignored — so the fallback flip carries no hardware risk.

### Findings

**1. `has_flow` is silently granted to every newly-routed variant — the one thing worth a second look.**

`BHyveHT25G2Device.has_flow = True` (`devices/ht25g2.py:32`), and the mesh class sets nothing. So `HT25A-0001` and `HT25-0001` now gain flow capability as a side effect of the routing change. That isn't cosmetic: it adds a `#57{#1=0}` unsubscribe send on *every* `refresh_status` (`protobuf.py:302`), flow reads on watering polls (`protobuf.py:467`), a flow sensor (`sensor.py:93`), a Check-flow button (`button.py:49`), and extra binary sensors (`binary_sensor.py:50`). The comment at `protobuf.py:115` explicitly says the flag exists "so we don't poke a flow-less XD" — this routes an unverified variant straight into the poking path.

@quadcom confirmed connect + actuate, not flow. It's plausibly correct for a 90205Z, but it's unverified. Could you ask on #47 whether the flow sensor populates on that unit? If it doesn't, we should split the flow capability out of the routing decision rather than let it ride along.

**2. No test for `HT25A-0001` — the actual string from #47.**

The added cases cover `HT25G2-0001/0098` and `HT25-0001/0098`, but the hardware string that motivated the whole fix isn't in the table. One line:

```python
("HT25A-0001", "0098", "", BHyveHT25G2Device),   # issue #47, HW-verified
```

**3. Signal priority is inverted — the strongest marker now loses.**

`endswith("-0000")` is checked *before* the explicit `HT25G2` prefix, so a hypothetical `HT25G2-0000` would route to mesh. Nobody has seen that string, but the explicit G2 marker is the more authoritative signal and it's free to keep it first:

```python
if (hardware or "").startswith("HT25G2"):
    return BHyveHT25G2Device
if (hardware or "").endswith("-0000"):
    ...
```

### Minor

- **The PR description overstates backward compat.** It says routing for `HT25-0001` is unchanged, but `HT25-0001` + fw `0085`/`0041` flips mesh → protobuf — the old tests asserted mesh for exactly those rows, and this PR rewrites them to `HT25-0000`. The change is almost certainly right (the docs treat `HT25-0001` as protobuf hardware, and those old rows look synthetic rather than observed), but "only the fallback for unrecognized firmware changes" isn't literally accurate. Worth correcting in the description so the next person reading the history isn't misled.
- **Stale comment:** `devices/ht25g2.py:8` still says the dispatcher "disambiguates by hardware-suffix / firmware". Firmware is gone from that branch now.
- **README support table** should gain the now-verified `HT25A-0001` / `0098` row.
- `log_label = "HT25G2"` means an `HT25A-0001` unit logs as `HT25G2`. Cosmetic — diagnostics reports `hardware` separately — but noting it.

Items 2 and 3 are one-liners; if you push those plus the README row I'll merge. Item 1 I'd like an answer on from #47 first, though it doesn't have to block this PR if the flow entities turn out to be harmless on that unit.

exactmike:
Pushed fixes for items 2 and 3, plus the README row:

- **Finding 3 (priority inversion):** `HT25G2` prefix is now checked before the `-0000` suffix, so a hypothetical `HT25G2-0000` would route to protobuf.
- **Finding 2 (missing test):** added `("HT25A-0001", "0098", "", BHyveHT25G2Device)` and an `HT25G2-0000` case covering the new priority order.
- Fixed the stale comment in `ht25g2.py:8` and added the `HT25A-0001`/`0098` row to the README support table.
- Corrected the PR description — the backward-compat claim was wrong about `HT25-0001` on `fw0085`/`fw0041`; updated to describe the actual (unobserved-combination) change.

CI is green on the new commit (7ff84e9).

For finding 1 (`has_flow`), I'll follow up on #47 to ask quadcom whether the flow sensor populates on his unit and report back here — agreed it shouldn't block merge if the entities turn out to be harmless, but worth confirming.

exactmike:
Follow-up from the PR notes: quadcom's flow sensor results are in on #47.

The primary "Flow rate" sensor (slope-derived from the raw counter) reports correctly on his `HT25A-0001`/fw0098 unit, confirming `has_flow` routing correctly extends to it as a side effect of this fix.

The diagnostic-only "Flow rate (device)" sensor (`#59.#4 currentFlowRateGpm`, disabled by default) does not populate on his unit — consistent with our own Gen2 fleet (fw0111), where it's also never been observed populated. Not a regression from this PR; it's pre-existing, disabled-by-default behavior from #39, and this is just further field confirmation that the field itself isn't sent by any firmware we've tested.

No changes needed here — just closing the loop.

Issue #52, opened 2026-09-01
Settings are not passed to the valve itself


quadcom:
I'm trying to set the rain delay, either by hour or using the presets, and any changes made to either aren't being passed to the valve itself. If I refresh the integration devices page, all the values I had set for those fields are reset back to 0 and OFF. The device shows as connected, and status updates show sync events are connecting.

This is happening on both the HT34A-0001 and the HT25A-0001.

I've looked at the HAOS core and supervisor logs but nothing under "Orbit" or "Hyve" is returned in any searches. Where else can I find log data to send you?

<img width="1458" height="736" alt="Image" src="https://github.com/user-attachments/assets/c7d3d469-18ca-4ab0-ab02-574fbfc26077" />

<img width="1464" height="674" alt="Image" src="https://github.com/user-attachments/assets/329d9f29-a7ac-488c-9d83-3499d825c36d" />

exactmike:
Thanks for the report and the screenshots. The firmware versions are useful. Your HT34A-0001 on fw 0107 is the same hardware and firmware as the XD I used to test rain delay when the preset select landed in v2.3.0. So the cause is most likely in the environment, not in the firmware. Two different failures produce the same symptom. The steps below tell them apart.

When you set a rain delay, the integration writes it to the device. It then reads the device status back and treats the delay as set only if the device echoes it. If the read-back does not show the delay, the entity reverts to 0 on the next refresh with no error. It must show an error instead, and I will fix that separately. So "reset to 0" means one of two things:

1. The device did not accept or store the delay. This is a real "not passed to the valve."
2. The device stored the delay, but Home Assistant discards it. This happens if the device clock is far behind the Home Assistant clock. The expiry is anchored to the device clock, so Home Assistant sees the delay as already expired.

To tell them apart, please do the following:

[...]

PR #53, opened 2026-09-12
fix(rain-delay): raise a service error when the device does not confirm the write

exactmike:
Refs #52.

### The bug

`set_rain_delay()` and `clear_rain_delay()` return `False` when the device does not echo
the new `#16.#13` state after the write. The "Rain delay" number and the "Rain delay
preset" select discarded that return value. An unconfirmed write looked like success in
the UI, and the entity went back to 0 on the next refresh. Nothing pointed the user at
the log. That is the experience that #52 describes.

A second gap showed up while I traced this. `set_rain_delay` and `clear_rain_delay` were
the only protobuf writes without `_ble_write_guard`. A transient `BleakError` on a
marginal proxy link surfaced as a raw exception from the entity call. Every other write
logs a WARNING and returns `False` in that case.

### The fix

Both entities now route through one helper, `number.async_apply_rain_delay()`. The
helper writes the delay, then requests a coordinator refresh in every case so that the
entity shows the real device state. If the device did not confirm the write, the helper
raises `HomeAssistantError`. The user gets the standard service error toast with the
device name and a pointer to the log.

`set_rain_delay` and `clear_rain_delay` now carry `_ble_write_guard`. A BLE error logs
the existing WARNING and returns `False`, like the other writes.

The behavior of a confirmed write is unchanged. The `valve.py` open and close calls and
the program services discard the same return value. That is left for a separate PR.

### Verification

Tests:

- `tests/test_devices.py::test_rain_delay_writes_return_false_on_ble_error` runs for set
  and for clear. It fails on the unguarded methods and passes with the guard.
- `tests/test_ha_rain_delay.py` needs Home Assistant. It covers the raise after refresh
  for set and for clear, and the silent confirmed path.
- 254 passed, 3 skipped locally. The Validate workflow (Hassfest, HACS, pytest with Home
  Assistant) is green on the fork.

Hardware (HA 2026.9, ESPHome proxies, branch deployed to `/config`):

- Gen2 `HT25G2-0001` fw0111: the "2 hours" preset logs `rain-delay set 120m confirmed`
  and the service returns normally. "Off" clears the delay. The confirmed path is
  unchanged.
- XD `HT34A-0001` fw0107: the device does not echo the same "2 hours" write. The log shows
  `rain-delay set 120m unconfirmed`. The service call now fails with
  `BT4ValveXD01: rain delay not confirmed by the device (see log)` instead of a silent
  revert. This reproduces the #52 symptom on our own unit, with Automatic watering on
  and off. The cause on the device side is tracked in #52 and is independent of this PR.


ljmerza (commented):
Looks correct. Nothing blocking. I ran the full suite locally on this branch with Home Assistant installed: 282 passed, 0 skipped.

**What looks right**
- The guard works on both writes. `set_rain_delay` and `_clear_rain_delay_unlocked` each disconnect in their own `try/finally` before `_ble_write_guard` catches the error, the same as the other guarded writes.
- Refreshing before raising is the right order: the entity shows the real device state before the error appears.
- The shared helper is equivalent to the old code. A number value that rounds to 0 minutes still clears the delay.

**Non-blocking**
1. **Misleading message when there is no connection.** `set_rain_delay` / `clear_rain_delay` return `False` when `self.connection is None` (`protobuf.py:587`, `:633`). The user then sees "rain delay not confirmed by the device", but no write was ever sent. Consider a neutral message, e.g. "rain delay write not confirmed (see log)".
2. **The error arrives after a full BLE poll.** The helper waits for `async_request_refresh()` before raising. The debouncer runs the first call immediately (see the note in `__init__.py`), so the error shows only after a complete status read over the proxy. The old code also waited for the refresh, so this is not a regression, just more noticeable now.
3. **Cross-platform import.** `select.py` now imports `async_apply_rain_delay` from `number.py`. There is no import cycle today, but a shared write helper fits better in `coordinator.py` or a small helpers module than in the number platform.
4. **XD echo behavior.** Comments in `set_rain_delay` note that the device suppresses the `#16.#13` echo while a delay is active. If the HT34A applies the delay without echoing it, this would raise an error for a write that succeeded. Your report that the entity reverts to 0 suggests the write really fails, so this is only something to keep in mind while #52 is investigated.
5. **Test gap.** The HA tests call the helper directly, and nothing checks that the number and select entities call it. The wiring is one line each, so that's acceptable.

**Out of scope, for the follow-up PR**
Besides `valve.py` and the program services, `switch.py` has the same silent failure: `set_controller_mode` (auto watering switch) and `set_program_enabled` (program A–D switches) also ignore the `bool` they return.

exactmike:
Thanks for the review. Pushed one commit for items 1 and 3:

1. The message is now `<device>: rain delay write not confirmed (see log)`, so it no longer claims the device answered when `connection is None` returned `False` before any write.
3. `async_apply_rain_delay` moved onto `BHyveDeviceCoordinator`. `number.py` and `select.py` call `self.coordinator.async_apply_rain_delay(...)`, and the cross-platform import is gone. The test drives the same method.

Items 2 and 5 left as they are, for the reasons you gave.

Item 4, from the hardware side: on my HT34A-0001 (fw 0107) the write does fail on the device. The status echo after every `#17` still carries a rain-delay block from a 3-day delay I set on Aug 22 that expired on its own on Aug 25 (`#1=4320`, `#3` in the past, `#4=1`), with run-state 1 and no `#5` remaining-minutes field. Since then the device refuses set and clear from Home Assistant in every `#17` field combination I tried, from its own front-panel Rain Delay button, and from the B-Hyve app, and the block survives a battery pull. The `#17` handler does answer with a `#16` (an unknown message gets no reply), it just leaves the record alone. My four HT25G2-0001 valves on the same network key accept the same write and confirm normally. So the raise is correct for this case, and I am following up in #52 on whether quadcom's units are in the same state.

Follow-up PR for `switch.py`, `valve.py`, and the program services noted.

ljmerza:
Merged, thanks. The same pattern for the switches is in #55: I moved the refresh-then-raise step into `BHyveDeviceCoordinator.async_confirm_write` and used it in the automatic-watering and program switches, so there's no need for a separate switch.py PR on your side. `valve.py` and the program services are still open if you want them.

